PR #4, opened 2016-10-28
fix lock ordering race condition in dial sync code

whyrusleeping:


whyrusleeping:
The issue was that a concurrent incref and decref could deadlock:

A finishes a dial call and calls decref, A takes the activedial Lock
B calls DialSync and takes the dialSync lock
A now has the activedial lock and attempts to take the dialSync lock
B now has the dialSync lock and attempts to take the activedial lock.
- pretty basic deadlock.

cc @Kubuxu and @lgierth for review

dial_sync.go
```diff
@@ -53,22 +54,38 @@ func (ad *activeDial) incref() {
...
+		ad.refCntLk.Lock()
+		// check again after lock swap drop to make sure nobody else called incref
+		// in between locks
+		if ad.refCnt <= 0 {
+			ad.cancel()
+			delete(ad.ds.dials, ad.id)
+		}
 		ad.ds.dialsLk.Unlock()
+		ad.refCntLk.Unlock()
 	}
```
kevina: Nit: I would release ad.refCntLk.Unlock() first, don't think it makes any difference in this case.

dial_sync.go
```diff
@@ -53,22 +54,38 @@ func (ad *activeDial) incref() {
 
 func (ad *activeDial) decref() {
 	ad.refCntLk.Lock()
-	defer ad.refCntLk.Unlock()
 	ad.refCnt--
```
kevina: Since you are no longer using defer anywhere in this function I am going to assume that nothing can panic otherwise the locks might never be released.

dial_sync.go
whyrusleeping: Ah yeah, good point.

dial_sync.go
kevina: So cancel() is guaranteed not to panic?

whyrusleeping:
@kevina i moved the incref back into the `getActiveDial` method to ensure we don't end up having an object that has been cleared already:

A holds an activeDial with refCnt = 1
B calls getActiveDial, gets the same activeDial that A has.
A returns and calls decref, reduces refCnt to 0 and then cancels its context.
B then calls incref and wait, but wait returns immediately and fails because its context has already been canceled.

dial_sync.go
whyrusleeping: Yeah, cancel won't panic

kevina:
@whyrusleeping, I'm afraid I don't understand this code well enough to give it a proper audit.  I don't see how this code can deadlock now, but there might still be race condensations.

Kubuxu:
LGTM, something like actor model for DialSync would make the code much safer and prevent bugs like that but I can see the tradeoffs.

I still don't like that we have more than one lock in play here but I will have to live with it.

whyrusleeping:
Yeah... switching to an actor model here complicates the code rather significantly. I normally write code that way, but it wasnt going to be fun this time around.

PR #8, opened 2016-11-14
Introduce changes required for Private Netowrk support

Kubuxu:


Kubuxu:
@whyrusleeping for CR

PR #11, opened 2016-11-24
Expose errors for private network

Kubuxu:
@lgierth it would be great if it was in before you run dep update.
@whyrusleeping for CR.


swarm_dial.go
```diff
@@ -158,7 +159,11 @@ func (s *Swarm) Dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		return nil, ErrDialToSelf
 	}
 
-	return s.gatedDialAttempt(ctx, p)
+	conn, err := s.gatedDialAttempt(ctx, p)
+	if s.protec != nil && strings.Contains(err.Error(), "privnet") {
+		log.Warninf("error: %s. Possible missmatched private network key", err.Error())
+	}
+	return conn, err
```
lgierth: It's `log.Warningf()` -- I think this is also a good case for error types in go-libp2p-interface-pnet. There is a dependency here, we're just hiding it away in a string.

lgierth:
> it would be great if it was in before you run dep update.

ok -- I aim to do the update thing tomorrow afternoon before @whyrusleeping wakes up. Otherwise we'll have plenty more opportunities for updates :)

swarm_dial.go
lgierth: You could even have a "root" error type in go-libp2p-interface-pnet, and more specific error types derived from that in the implementation package(s).

Kubuxu: @lgierth how do you like this: https://github.com/libp2p/go-libp2p-interface-pnet/blob/master/error.go

lgierth: I think it can be simplified to just `type Error error` and `type SomeError ifacepnet.Error`? Then you can just type-switch if you need to know

Kubuxu: This won't work as I won't be able to detect if SomeError is ifacepnet.Error unless I am missing something.

Kubuxu: Or maybe I will be, testing.

Kubuxu: No as type assertion for type alias passes without any problem.

`_, ok := errors.New("generic error").(ifacepnet.Error)` would have ok = true.

Kubuxu:
Updated it to use new ipnet but it still needs updated ipnet.

whyrusleeping:
@Kubuxu status here? I think this change LGTM. Needs a rebase though

Stebalien:
This feels like a weird place to log this. Why not just log in `go-libp2p-pnet` when we encounter the error?

PR #16, opened 2017-01-24
Add multiplex transport

VictorBjelkholm:


Stebalien:
Currently, we do this at the application level (and call the transport `mplex`, that is, `/mplex/6.7.0`). @whyrusleeping do we want to do this here instead?

Stebalien:
Note: Travis is breaking because multiplex needs to be imported as a gx dep.

Kubuxu:
I would go onto something else, can we remove those transports form here and have the app add them? This way the app has freedom to choose and we might in future reduce our dependency tree.

As we are planning transport plugins, it might be wise to make them top level injected.
Only other packages depending on muxers are (looking at yamux, the oldest one):
1. go-libp2p-swarm - which is using them in the same exact manner as this package
2. go-peerstream
3. go-stream-muxer - for tests, probably can be swapped out for some dummy muxer

whyrusleeping:
Closing this, we're moving away from embedding *any* multiplexers here.

PR #19, opened 2017-03-17
Add Swarm.Transports() getter

lgierth:
This is so that the gateway can do some casting a la `n.PeerHost.Network().(*swarm.Network).Swarm()` and then get the transports, in order to grab the /ws transport, and mount its listener into the gateway http muxer.

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -183,6 +183,10 @@ func (s *Swarm) AddTransport(t transport.Transport) {
 	s.transports = append(s.transports, t)
 }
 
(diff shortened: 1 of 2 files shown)


lgierth:
Mh I think what I actually want is grabbing the ws Listener from the peerstream swarm...

Kubuxu:
I think it isn't good idea, the gateways should have the listener exposed directly.

Create the WS transport, put it into supported transports and save reference for gateway.

lgierth:
Yeah I think so too... nevermind this PR :)

PR #21, opened 2017-04-17
use new conn interfaces

marten-seemann:
This PR depends on https://github.com/libp2p/go-libp2p-transport/pull/20.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -373,22 +373,9 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (ico
 	return connC, nil
 }
 
-var ConnSetupTimeout = time.Minute * 5
-
 // dialConnSetup is the setup logic for a connection from the dial side. it
 // needs to add the Conn to the StreamSwarm, then run newConnSetup
 func dialConnSetup(ctx context.Context, s *Swarm, connC iconn.Conn) (*Conn, error) {
-	// TODO: do we still need this?
-	// deadline, ok := ctx.Deadline()
-	// if !ok {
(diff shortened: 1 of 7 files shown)


swarm_dial.go
```diff
@@ -377,15 +378,16 @@ var ConnSetupTimeout = time.Minute * 5
 // dialConnSetup is the setup logic for a connection from the dial side. it
 // needs to add the Conn to the StreamSwarm, then run newConnSetup
 func dialConnSetup(ctx context.Context, s *Swarm, connC iconn.Conn) (*Conn, error) {
```
marten-seemann: @whyrusleeping: I'm not sure what the best solution for the deadline here. `SetDeadline` doesn't make sense for a stream-multiplexed connection (QUIC supports deadlines for streams as soon as https://github.com/lucas-clemente/quic-go/issues/514 is fixed, but I don't think this will help us here).

The call to `peerstream.AddConn` shouldn't take any time, so we only have to worry about the `newConnSetup` call. We're already passing the context there, so we could do try to implement there.
What do you think?

whyrusleeping:
The thing here is that the multistream negotiation for which stream muxer
to use happens in peerstream.AddConn ( I think). That's why there's a set
timeout. You're right that it doesn't make sense for a connection that
already has stream multiplexing, we can probably move the deadline stuff
down as you suggest

marten-seemann:
We don't need it here at all. The handshake is performed in go-libp2p-conn, so none all function calls in `newConnSetup` should return immediately. I removed the deadlines there.

whyrusleeping:
@marten-seemann sounds good

whyrusleeping:
I think we can close this one now, right?

Stebalien:
Yes.

PR #25, opened 2017-07-20
Release 1.7.0

Stebalien:
And fix tests for the synchronous notification changes.


whyrusleeping:
for the sake of these tests, could you not just add a buffer to the channels?

vyzo:
@Stebalien any progress? I am blocked on this for go-libp2p-circuit.

Stebalien:
> for the sake of these tests, could you not just add a buffer to the channels?

Done. We discussed alternative solutions offline but they all caused problems because we'd stop reading from the notification channels before `connectSwarm` was done writing to them.

PR #26, opened 2017-07-28
Fix dialLimiter.fdConsuming counting

magik6k:
This fixes https://github.com/ipfs/go-ipfs/issues/4102

With this fix when running daemon SYN_SENT connections are hovering around 150-200, which is a huge improvement from around 3000.

Unit tests are TODO

### DIFF
--- a/limiter_test.go
+++ b/limiter_test.go
@@ -320,3 +320,69 @@ func TestStressLimiter(t *testing.T) {
 		}
 	}
 }
+
+func TestFDLimitUnderflow(t *testing.T) {
+	dials := 0
+
+	df := func(ctx context.Context, p peer.ID, a ma.Multiaddr) (iconn.Conn, error) {
+		dials++
+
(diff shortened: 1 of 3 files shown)


magik6k:
@whyrusleeping Fixed the fix to actually fix what needed to be fixed. +Added a test for it

limiter_test.go
```diff
@@ -320,3 +320,70 @@ func TestStressLimiter(t *testing.T) {
...
+			}
+		}(peer.ID(fmt.Sprintf("testpeer%d", i % 20)), i)
+	}
+
+	time.Sleep(time.Second * 3)
+
+	if l.fdConsuming < 0 {
+		t.Fatalf("l.fdConsuming < 0")
+	}
+	fmt.Println(dials)
```
Kubuxu: Best to remove the log here or use `t.Log`

magik6k: Didn't notice when doing commit, removed

PR #33, opened 2017-08-31
test: connect each peer precisely once.

Stebalien:
Otherwise, some tests will get confused by duplicate (temporary, we close duplicates quickly) connections.

fixes libp2p/go-libp2p#100


Stebalien:
Meh. Reviews are for people who make mistakes, I *never* make mistakes. NEVER! (but really, this isn't even worth an LGTM).

PR #40, opened 2017-11-10
don't pass the dial timeout option

Stebalien:
It doesn't exist anymore. This is now handled entirely though contexts.

### DIFF
--- a/.gx/lastpubver
+++ b/.gx/lastpubver
@@ -1,1 +1,1 @@
-2.0.5: QmdQFrFnPrKRQtpeHKjZ3cVNwxmGKKS2TvhJTuN9C9yduh
+2.1.0: QmcnHXGwX6Yfqm7guiAjCujXg7RugKSaV1wj8RH2NTpNKK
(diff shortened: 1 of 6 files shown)


Stebalien:
Depends on libp2p/go-libp2p-conn#18 (will need some gx updates).

Stebalien:
This is failing because of a really weird bug that only triggers on Travis. By really weird, I mean I should be impossible. For some reason, it's failing to recognize that `/ip6/::/tcp/xxx` is an unspecified address.

Any help here would be appreciated, I can't reproduce locally.

Note: This bug also shows up in previous builds (pick an old build and try triggering a rebuild). Travis changed something. However, it still looks like our bug (probably in go-addr-util or go-multiaddr or go-multiaddr-net).

We can probably just go ahead and merge this (after testing locally) but this is infuriating.

Stebalien:
So, I missed a `!` when reading the go-multiaddr-net code. This is still funny but I'm going to merge anyways.

Issue filed as #42.

PR #45, opened 2017-11-21
nil out references to finished dials

Stebalien:
We need a better solution (this is a memory hotspot when mass dialing) but that
will take some thought and effort. This is a simple fix that should reduce
memory usage a bit.

### DIFF
--- a/limiter.go
+++ b/limiter.go
@@ -71,6 +71,7 @@ func (dl *dialLimiter) finishedDial(dj *dialJob) {
 
 		if len(dl.waitingOnFd) > 0 {
 			next := dl.waitingOnFd[0]
+			dl.waitingOnFd[0] = nil // clear out memory
 			dl.waitingOnFd = dl.waitingOnFd[1:]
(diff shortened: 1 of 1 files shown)


limiter.go
```diff
@@ -71,6 +71,7 @@ func (dl *dialLimiter) finishedDial(dj *dialJob) {
 
 		if len(dl.waitingOnFd) > 0 {
 			next := dl.waitingOnFd[0]
+			dl.waitingOnFd[0] = nil // clear out memory
```
Stebalien: This is really the main issue. If we build a massive backlog and then stop dialing, we won't free *anything* till we're done clearing it unless we do this.

Kubuxu: Is it because inaccessible objects from re-sliced slice are sill alive according to GC?

I didn't think this would be a case.

Stebalien:
Yes, unless the slice is reallocated (as far as I know).

https://github.com/golang/go/wiki/SliceTricks

PR #48, opened 2017-12-09
Remove check that we have the right peer on dial

jackkleeman:
This check now occurs in Dialer.Dial:
https://github.com/libp2p/go-libp2p/issues/137
https://github.com/libp2p/go-libp2p-conn/pull/22


Stebalien:
Thanks. I'll merge this after I update the secio gx dep.

Stebalien:
We decided to keep checking the peer ID for sanity reasons. It's fast and can't hurt.

PR #51, opened 2018-01-03
clear out extra dial jobs after dial finishes

whyrusleeping:
This really wouldnt be an issue if we didnt have tens of thousands of addresses hanging around for some reason

### DIFF
--- a/limiter.go
+++ b/limiter.go
@@ -144,6 +144,15 @@ func (dl *dialLimiter) AddDialJob(dj *dialJob) {
 	go dl.executeDial(dj)
 }
 
+func (dl *dialLimiter) clearAllPeerDials(p peer.ID) {
+	dl.rllock.Lock()
+	defer dl.rllock.Unlock()
+	delete(dl.waitingOnPeerLimit, p)
+	// NB: the waitingOnFd list doesnt need to be cleaned out here, we will
+	// remove them as we encounter them because they are 'cancelled' at this
+	// point
+}
+
 // executeDial calls the dialFunc, and reports the result through the response
 // channel when finished. Once the response is sent it also releases all tokens
 // it held during the dial.
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -293,6 +293,8 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
 	defaultDialFail := fmt.Errorf("failed to dial %s (default failure)", p)
 	exitErr := defaultDialFail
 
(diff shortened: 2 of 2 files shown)


whyrusleeping:
@Stebalien how does it not solve the problem?

limiter.go
```diff
@@ -144,6 +144,20 @@ func (dl *dialLimiter) AddDialJob(dj *dialJob) {
...
 }
 
+func (dl *dialLimiter) clearAllPeerDials(p peer.ID) {
+	dl.rllock.Lock()
+	defer dl.rllock.Unlock()
+	delete(dl.waitingOnPeerLimit, p)
+
+	arr := dl.waitingOnFd[:0]
+	for _, j := range dl.waitingOnFd {
+		if j.peer != p {
```
Stebalien: These are currently considered "active". We'll need to decrement the activePerPeer counter for each one we remove.

limiter.go
```diff
@@ -144,6 +144,15 @@ func (dl *dialLimiter) AddDialJob(dj *dialJob) {
 	go dl.executeDial(dj)
 }
 
+func (dl *dialLimiter) clearAllPeerDials(p peer.ID) {
+	dl.rllock.Lock()
+	defer dl.rllock.Unlock()
+	delete(dl.waitingOnPeerLimit, p)
+	// NB: the waitingOnFd list doesnt need to be cleaned out here, we will
+	// remove them as we encounter them because they are 'cancelled' at this
+	// point
```
Stebalien: Same as the `waitingOnPeerLimit` list. The problem is that it could be a *while* before that happens if we're really backed up behind a bunch of slow dials. It won't be *quite* as bad because it's limited by `dials*perPeerLimit` but it will still be large and can still grow in the same way.

whyrusleeping: @Stebalien Yeah, the tradeoff we make is the O(n) operation under a held lock after completion  vs this potential extra memory buffering.

whyrusleeping:
So the final thing I need to do here is add metrics as @lgierth requested

whyrusleeping:
alright, so it turns out we don't have things set up yet to nicely add metrics from go-libp2p. I'm thinking that we should push to get this merged sooner, and work out the metrics later

lgierth:
> I'm thinking that we should push to get this merged sooner, and work out the metrics later

Agreed

limiter.go
Stebalien: That could be solved with a doubly linked list (wait queues like this are really one of the only uses for them).

whyrusleeping: Ah, great point. We never really need to iterate over it, so that makes perfect sense.

PR #54, opened 2018-01-09
cleanup connection setup

Stebalien:
We already setup the connection from within the new connection handler. No need to do it *again* on dial.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -189,14 +189,16 @@ func (s *Swarm) gatedDialAttempt(ctx context.Context, p peer.ID) (*Conn, error) 
 // doDial is an ugly shim method to retain all the logging and backoff logic
 // of the old dialsync code
 func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
-	var logdial = lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
+	ctx, cancel := context.WithTimeout(ctx, s.dialT)
+	defer cancel()
+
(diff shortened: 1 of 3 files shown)


swarm_dial.go
```diff
@@ -189,14 +189,16 @@ func (s *Swarm) gatedDialAttempt(ctx context.Context, p peer.ID) (*Conn, error)
 // doDial is an ugly shim method to retain all the logging and backoff logic
 // of the old dialsync code
 func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
-	var logdial = lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
+	ctx, cancel := context.WithTimeout(ctx, s.dialT)
```
Stebalien: More idiomatic.

swarm_dial.go
```diff
@@ -206,8 +208,6 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
 		// ok, we failed. try again. (if loop is done, our error is output)
 		return nil, fmt.Errorf("dial attempt failed: %s", err)
 	}
-	log.Event(ctx, "swarmDialBackoffClear", logdial)
-	s.backf.Clear(p) // okay, no longer need to backoff
```
Stebalien: We handle this from within the the new connection handler.

whyrusleeping:
My only comment here would be that we are moving more logic into connHandler, which is supposedly called while holding a lock.

Stebalien:
This was already happening within `connHandler`, we were just doing it twice.

PR #57, opened 2018-02-15
[RFC] Delay dials to realy addresses

magik6k:
This is a somewhat naive way to do this, but I can't think of an easy way to make this smarter without making this ugly in some way.

One way this could be improved would be to switch from chan to an array in `dialAddrs`, I just wasn't sure what is the reason behind it being a channel, so I decided to not touch that for now.


Stebalien:
I see the issue but I'd like to discuss this thoroughly first (in an issue). More generally, we need some way to prefer some transports over others.

Note: We're using a channel because, *technically*, we'd like to dial and discover addresses in parallel. However, we don't currently do that. This does make things a bit annoying... but we can always try reading off what's on the channel into an array and then sorting there. We can even, e.g., introduce delays if we have different "preference" tiers.

whyrusleeping:
Yeah, ideally we are able to discover new addresses and inform the dial process of them during a dial. I did the refactor to move towards that, but the other side (the dht address finder) hasnt caught up yet. This is a hard problem.

magik6k:
> so what happens if a peer only has a relay address?
shouldn't we dial that immediately instead of waiting for 2s?

That's the hard part

Stebalien:
> That's the hard part

Expanding on my previous comment, I think the better (although certainly not perfect) solution here is to:

1. Read off the channel until it blocks (using a select). Maybe with a timeout (e.g., wait until it blocks for more than a millisecond).
2. Sort that slice of addresses into preference tiers.
3. Add delays between preference tiers (iff there exists an address in the preference tier).

magik6k:
Ok, I think I pushed something that makes sense, please review.

todo:
- tests
- a bit more abstracted `dialOffset`
  - `dialLimiter` doesn't exactly abstract it's checks.
  - Should I PR p2p-circuit entry to go-multiaddr and to mafmt?

whyrusleeping:
cc @diasdavid. Something something relay dialing logic.

magik6k:
Refactored to use buckets

dial_delay.go
```diff
@@ -0,0 +1,167 @@
...
+	"time"
+
+	ma "github.com/multiformats/go-multiaddr"
+	mafmt "github.com/whyrusleeping/mafmt"
+)
+
+const p_circuit = 290
+
+const numTiers = 2
+const tierDelay = 2 * time.Second
```
vyzo: can we make it just 1s? 2s might already be too much for a dial.

dial_delay.go
```diff
@@ -0,0 +1,167 @@
...
+					if !ok {
+						break outer
+					}
+					put(addr)
+				default:
+					break fill
+				}
+			}
+
+			next, tier := get()
```
magik6k: Note that with the current setup we don't ever reach this point in this loop as the channel is closed by now. All sending will happen in the loop in L127. I'll add some tests later today to test that it works.

Stebalien:
With a test, LGTM. You even covered the "out of addresses" case. Nice!

(I kind of expected you to tell me to tell me my idea was way too complicated and to take a hike...)

dial_delay.go
```diff
@@ -0,0 +1,167 @@
...
+
+// delayDialAddrs returns a address channel sorted by priority, pushing the
+// addresses with delay between them. The other channel can be used to trigger
+// sending more addresses in case all previous failed
+func delayDialAddrs(ctx context.Context, c <-chan ma.Multiaddr) (<-chan ma.Multiaddr, chan<- struct{}) {
+	out := make(chan ma.Multiaddr)
+	delay := time.NewTimer(tierDelay)
+	triggerNext := make(chan struct{}, 1)
+
+	go func() {
```
whyrusleeping: would be nice to have this broken out into a separate function

dial_delay.go
```diff
@@ -0,0 +1,167 @@
...
+		lastTier := 0
+
+		// put enqueues the mutliaddr
+		put := func(addr ma.Multiaddr) {
+			tier := getTier(addr)
+			pending[tier] = append(pending[tier], addr)
+		}
+
+		// get gets the best (lowest tier) multiaddr available
+		get := func() (ma.Multiaddr, int) {
```
whyrusleeping: worth noting that we're using a stack within each tier. Might not be the best approach (though definitely simpler and cheaper to implement)

magik6k:
* Broken the code into some more functions
* Added tests for all cases I could imagine
  * all cases covered
  * ran the tests 20 times, none have failed (not even on windows)

whyrusleeping:
I would also like @diasdavid to review this. This is a libp2p design decision I think he should have a say in.

dial_delay.go
```diff
@@ -0,0 +1,210 @@
...
+	delay := time.NewTimer(tierDelay)
+	triggerNext := make(chan struct{}, 1)
+
+	go func() {
+		defer delay.Stop()
+		defer close(out)
+		var pending [numTiers][]ma.Multiaddr
+		lastTier := -1
+
+		// put enqueues the mutliaddr
```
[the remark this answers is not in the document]
magik6k: ?

vyzo:
I have a real-world concern with this mechanism.

It is implied that the dialing of the relay address won't start until the previous dials have failed.
However, in the case of NAT, the primary failure mode is to timeout (there is no RST from the firewall, the SYN just goes to a black hole). And these timeouts can be quite long.

The problem here is that when we introduce advertised relay addresses to overcome NAT connectivity issues, we might simply fail to dial them altogether because the previous tier doesn't fail in a timely fashion.
We need to make sure that this is addressed, or this patch will end up breaking things.

magik6k:
Dials through relay will start no later than after `tierDelay` since start of the dial. We could try to get some stats in the runtime to tune the variable to some lower value (e.g 90th% percent slowest successful dial for a given tier, or some const value, whichever would be lower - should be the delay for the tier).

But I'd leave that for another PR, this one has enough complexity already. The 2s value is chosen arbitrarily, it would be great to get some dial times and see what a dial usually takes (note that we don't log events for transport dials, only for the entire dial). Without having the data my guess for a conservative enough delay (assuming usage within earth, using low-end broadband connection) would be somewhere around 200-300ms.

vyzo:
Ok, good -- I had a nightmare last night that we deployed relays and they didn't work because we wouldn't dial them :)

I think we should lower the delay to 1s tops, as a successful direct dial should complete well under that.
Something like 5RTTs -> 1s for networks on which go-ipfs nodes typically run seems like ample time.

vyzo:
Btw, we so happen to have dial data from the [crawler](https://github.com/vyzo/ipfs-crawl).

Let me take a look at the distribution of successful dial times.

whyrusleeping:
@vyzo can you grab a bunch of the individual dial data (from the eventlogs) and see how long 'successful' dials take on average? We probably want to wait that long plus a standard deviation or two.

vyzo:
Yes, that's exactly what I am doing -- filter successful dials to get a distribution.

vyzo:
Dial times from 15570 successful dials:
https://gist.github.com/vyzo/688f95706d504493925543fd954e508f

will get the distribution too.

vyzo:
Note that my network vantage point is a home network node in .gr.

whyrusleeping:
Thats such great data. Could we get the raw capture as well? (including failures)

vyzo:
Sure, I'll pin the dial log from the last crawl.

vyzo:
Full dial event log here: ipfs.io/ipfs/QmU3CtjYS12wUjkczBUnuRrj6CzxKLxpoVTb2ZNZ8wFaDH

vyzo:
Working a bit more with the data, the dials reached 1089 peers in total, so there are multiple dial events peer peer (non uniform).
I'll extract the median dial time for each peer and work on the distribution of those 1089 datapoints.

vyzo:
Processed data (in ms):
- Median successful dial time: 
  - times: https://gist.github.com/vyzo/c0ebd4072305a1079b33de8da01b3e20
  - CDF: https://gist.github.com/vyzo/2438e925a4981993d2787847b5e93f73
- Minimum successful dial time: 
  - times: https://gist.github.com/vyzo/99c9766e759e6dc9e4be9b4b2ff9607e
  - CDF: https://gist.github.com/vyzo/32f31c543629a68dea6fd782ebfea207
- Maximum successful dial time:
  - times: https://gist.github.com/vyzo/090fce904d594df6c20ed09f54d7281a
  - CDF: https://gist.github.com/vyzo/45fd915076f0a5a440b7b076b9bc587c

CDF plot: 
  - https://ipfs.io/ipfs/QmWLJBihx1GkGxJMuYPos31UpXf3VZNjmWxmfPFwUTUwss

vyzo:
One might say 1s would cover 80% of successful dials and 2s 90% of them; I think peers with longer dials would probably benefit from relayed connections.

dial_delay.go
```diff
@@ -0,0 +1,208 @@
...
+
+	ma "github.com/multiformats/go-multiaddr"
+	mafmt "github.com/whyrusleeping/mafmt"
+)
+
+const p_circuit = 290
+
+const numTiers = 2
+
+var tierDelay = 2 * time.Second
```
whyrusleeping: make this exported, we will want to be able to tweak it easily i think

dial_delay.go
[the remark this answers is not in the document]
whyrusleeping: s/mutliaddr/multiaddr/

dial_delay.go
```diff
@@ -0,0 +1,208 @@
...
+						return false
+					}
+					put(addr)
+				default:
+					return true
+				}
+			}
+		}
+
+		// waitForMore woits for addresses from the channel
```
whyrusleeping: s/woits/waits/

diasdavid:
Thanks for the nudge, @whyrusleeping. Given that this is a libp2p design decision, can we make sure it happens through the libp2p/Relay Specs -- https://github.com/libp2p/specs -- so that all the implementations conform? It is about time :) //cc @mgoelzer @tomaka

As far as my take on this. In JS land we currently only dial to relay after all the others have failed for the same issues described in this PR. The setTimeout of 1 or 2 seconds is a clever idea. 

It would be good to store the info on what was the last transport used to dial to a peer, so that 2 consequent dials do not force a libp2p node to go through the guessing game again.

dial_delay.go
magik6k: done

swarm_dial.go
```diff
@@ -16,16 +16,39 @@ import (
...
+//                                                                      |||||
+//                                       eventually a connection will   |||||
+//                                        get established (C), other -> |C|||
+//       Connection setup will be        attempts will get cancelled    x|xxx
+//     performed with dialConnSetup.                                     |
+//     It adds the connection to the ----------------------------------> |
+//        swarm and agrees on the                                        |
+//            muxer to use                                             /---\
+//                                        The result is distributed -> |||||
+//                                            to callers of Dial       |||||
```
magik6k: @Stebalien after the transport refactor only the part after 'dialConnSetup' changed, right?

dial_delay.go
```diff
@@ -0,0 +1,209 @@
...
+	mafmt "github.com/whyrusleeping/mafmt"
+)
+
+const p_circuit = 290
+
+const numTiers = 2
+
+var TierDelay = 1 * time.Second
+
+var relay = mafmt.Or(mafmt.And(mafmt.Base(p_circuit), mafmt.Base(ma.P_IPFS)), mafmt.And(mafmt.Base(ma.P_IPFS), mafmt.Base(p_circuit), mafmt.Base(ma.P_IPFS)))
```
magik6k: I wonder if we could have the transports have some way to indicate that they should be delayed, possibly an optional interface like
```
type ExpensiveTransport interface {
  TransportCost() int //basically the 'tier' to use here
}
```

swarm_dial.go
Stebalien: Yes.

bigs:
per @diasdavid's suggestion should we close this temporarily and move to an RFC?

Stebalien:
I really can't remember what this was blocked on other than "holy shit this is complicated".

vyzo:
Reopening this, as it's a high priority issue.

vyzo:
So where are we blocked here?

magik6k:
https://github.com/libp2p/go-libp2p-swarm/pull/57#issuecomment-418547294

> holy shit this is complicated

No, but seriously, IIRC the core logic here is done, one thing that would be nice would be to have some mentod in the transport interface saying that the transport is lower tier so we don't hardcode relay here - https://github.com/libp2p/go-libp2p-swarm/pull/57/files#diff-46c472072de3b864e2fabb74ab784758R202, but I think this can by done separate to this PR

dial_delay.go
```diff
@@ -0,0 +1,209 @@
...
+
+import (
+	"context"
+	"time"
+
+	ma "github.com/multiformats/go-multiaddr"
+	mafmt "github.com/whyrusleeping/mafmt"
+)
+
+const p_circuit = 290
```
vyzo: we should probably capitalize this to `P_CIRCUIT` -- that's what we do everywhere else.

dial_delay.go
vyzo: We should be aware of potential problems with changes in the circuit addressing here -- cf https://github.com/libp2p/specs/issues/72

vyzo:
it's still quite complicated, but it looks ok; will reread in the morning.

@Stebalien care for a refresher review? Let's move it forward.

dial_delay.go
vyzo: This might break with the changes in https://github.com/libp2p/go-libp2p-circuit/pull/48
We should drop the trailing /ipfs part from the filter and just rely on the presence of /p2p-circuit

vyzo: Also problematic for explicit relay addresses.

dial_delay.go
vyzo: will do.

vyzo: done.

vyzo:
- Fixed issues with the relay address filter and future-proofed it for https://github.com/libp2p/go-libp2p-circuit/pull/48 and https://github.com/libp2p/specs/issues/72
- Broke out the delay logic to its own top-level function, per @whyrusleeping's suggestion.

PR #63, opened 2018-03-13
refactor for new transports

Stebalien:
Removes:

* go-libp2p-conn logic (moved to transports)
* go-peerstream logic (moved here)

Big Changes:

* New transport interface.
* Explicit over implicit (the implicit automagic now all lives in go-libp2p):
  * No more default transports, muxers, etc.
  * No more fallback dialer. Transports are now *required*.
* Stream opening:
  * Connection picking logic (instead of just picking the first).
  * Tries harder to open a stream of some connections happen to be closed.
* Stream closing:
  * No longer treats half-closed streams as fully closed. Users *must* read the an EOF or reset the stream for it to be garbage collected.
* No more polling for dead connections.
* Multiplexers are now *non-optional*. Really, they haven't been optional for a while but we still pretended that they were.
* No more Network type alias. It added a bunch of code and didn't really provide anything but an alternative set of methods that do the same thing.
* Notifications:
  * New guarantee: connection open notifications will complete before connection close notifications begin.
  * Given that, notifications are now delivered in parallel. No more notification backlogs blocking connection closing/opening.

DO NOT MERGE (yet)
Part of: libp2p/go-libp2p#297

This will not currently build (gx deps not updated). However, it works and passes the tests locally.


swarm_addr_test.go
```diff
@@ -1,100 +1,14 @@
...
 	"testing"
 
-	addrutil "github.com/libp2p/go-addr-util"
-	metrics "github.com/libp2p/go-libp2p-metrics"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 	testutil "github.com/libp2p/go-testutil"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-func TestFilterAddrs(t *testing.T) {
```
Stebalien: The swarm itself now filters addresses based on the registered transports.

swarm_dial.go
```diff
@@ -35,18 +38,24 @@ var (
...
+	// given multiaddr.
+	ErrNoTransport = errors.New("no transport for protocol")
 )
 
-// dialAttempts governs how many times a goroutine will try to dial a given peer.
+// DialAttempts governs how many times a goroutine will try to dial a given peer.
 // Note: this is down to one, as we have _too many dials_ atm. To add back in,
 // add loop back in Dial(.)
-const dialAttempts = 1
+const DialAttempts = 1
```
Stebalien: Tests needed access to these and, really, there's no reason not to export them. They're currently constants but we can make them variables later if we want them to be configurable (or even make them swarm parameters).

swarm_dial.go
```diff
@@ -74,7 +83,12 @@ const defaultPerPeerRateLimit = 8
 //  }
 //
 
-type dialbackoff struct {
+// DialBackoff is a type for tracking peer dial backoffs.
+//
+// * It's safe to use it's zero value.
+// * It's thread-safe.
+// * It's *not* safe to move this type after using.
+type DialBackoff struct {
```
Stebalien: Exported so we stop exporting a private type from a public function.

swarm_dial.go
```diff
@@ -104,75 +118,81 @@ func (db *dialbackoff) Backoff(p peer.ID) (backoff bool) {
...
-			until: time.Now().Add(baseBackoffTime),
+			until: time.Now().Add(BackoffBase),
 		}
 		return
 	}
 
-	expTimeAdd := time.Second * time.Duration(bp.tries*bp.tries)
-	if expTimeAdd > maxBackoffTime {
-		expTimeAdd = maxBackoffTime
+	backoffTime := BackoffBase + BackoffCoef*time.Duration(bp.tries*bp.tries)
```
Stebalien: Nit: Adding in the BackoffBase *here* ensures that we don't go over the maximum (we used to go over by BackoffBase).

swarm_dial.go
```diff
@@ -249,8 +296,10 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 	*/
 	paddrs := s.peers.Addrs(p)
 	goodAddrs := addrutil.FilterAddrs(paddrs,
-		addrutil.AddrUsableFunc,
 		subtractFilter,
+		s.canDial,
+		// TODO: Consider allowing this?
```
Stebalien: Eventually, I'd like to allow link-local stuff. For now, this preserves existing behaviour.

swarm.go
```diff
@@ -236,94 +246,180 @@ func (s *Swarm) Close() error {
...
+	default:
+		errCh := make(chan error)
+		for _, c := range conns {
+			go func(c inet.Conn) {
+				errCh <- c.Close()
+			}(c)
 		}
 
-		seen[p] = struct{}{}
+		var errs []string
```
Stebalien: Is there a better way to do this? This kind of sucks.

Stebalien: Try templates + a wait group.

swarm_stream.go
```diff
@@ -1,69 +1,165 @@
...
-// Stream is a wrapper around a ps.Stream that exposes a way to get
-// our Conn and Swarm (instead of just the ps.Conn and ps.Swarm)
-type Stream ps.Stream
+type streamState int
 
-// Stream returns the underlying peerstream.Stream
-func (s *Stream) Stream() *ps.Stream {
-	return (*ps.Stream)(s)
+const (
+	streamOpen = iota
```
whyrusleeping: probs want to use `streamState(iota)`

swarm_transport.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, p := range protocols {
+		transport, ok := s.transports.m[p.Code]
+		if !ok {
+			continue
+		}
+		if transport.Proxy() {
+			return transport
+		}
+	}
+	p := protocols[len(protocols)-1]
```
magik6k: Is there any possibility that a multiaddress can specify no protocols? If yes, this may panic.

swarm_transport.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	transport "github.com/libp2p/go-libp2p-transport"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// TransportForAddr retrieves the appropriate transport for the given multiaddr.
```
magik6k: Would be nice to note that this implements the `Network` interface from `go-libp2p-transport` as it has nice docs

swarm_transport.go
magik6k: (yes, it would be very weird that such multiaddr would pop-up here but still)

Stebalien: There's a check for that at the top of the function (although it shouldn't happen).

swarm_transport.go
Stebalien: Fair enough.

swarm_stream.go
Stebalien: Done.

swarm_transport.go
```diff
@@ -26,8 +27,31 @@ func (s *Swarm) TransportForAddr(a ma.Multiaddr) transport.Transport {
...
+// the given multiaddr.
+func (s *Swarm) TransportForListening(a ma.Multiaddr) transport.Transport {
+	protocols := a.Protocols()
+	if len(protocols) == 0 {
+		return nil
+	}
+
+	s.transports.RLock()
+	defer s.transports.RUnlock()
+	selected := s.transports.m[protocols[len(protocols)-1].Code]
```
vyzo: will this work for TCP? don't we have `P_TCP` followed by `P_IP` there?

Stebalien: Other way around. We have IP followed by TCP. The idea is that, except for proxies like p2p-circuit, each multiaddr protocol "wraps" the previous one. So we pick the last one and then let it deal with the previous ones recursively (possibly delegating back to the swarm or some other transport manager). I'd rather build up from left to right but that starts to be come an abstraction nightmare.

> will

FYI, all this code *does* actually work in practice. I've been running go-ipfs with these changes (all the refactor code) for a while with no issue.

vyzo: Yeah, it looks good overall; just getting a little itchy with the possibility of introducing some very subtle bug.

Stebalien: I agree. And I know these are a *lot* of changes. That's why I'm letting them bake all together for a while while they're reviewed instead of merging them piece-wise.

dial_sync_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"time"
 
 	peer "github.com/libp2p/go-libp2p-peer"
+
```
Kubuxu: any reason for a new line here?

swarm.go
```diff
@@ -4,187 +4,169 @@ package swarm
...
 	peers pstore.Peerstore
-	connh ConnHandler
 
-	dsync *DialSync
-	backf dialbackoff
-	dialT time.Duration // mainly for tests
+	conns struct {
+		sync.RWMutex
+		m map[peer.ID][]*Conn
+	}
```
Kubuxu: I love this pattern.

swarm.go
```diff
@@ -4,187 +4,169 @@ package swarm
...
+			sync.RWMutex
+			m map[transport.Listener]struct{}
+		}{
+			m: make(map[transport.Listener]struct{}),
+		},
+		transports: struct {
+			sync.RWMutex
+			m map[int]transport.Transport
+		}{
+			m: make(map[int]transport.Transport),
```
Kubuxu: I wonder if it would be better to instead do something like this:
```
s.conns.m =  make(map[peer.ID][]*Conn)
```
and so on.
This way we don't have to define those types.

swarm.go
Kubuxu: Yeah, wait group would be much nicer pattern here.

swarm_conn.go
```diff
@@ -1,122 +1,220 @@
...
+func newConn(s *Swarm, c transport.Conn) *Conn {
+	return &Conn{
+		conn:  c,
+		swarm: s,
+		streams: struct {
+			sync.Mutex
+			m map[*Stream]struct{}
+		}{
+			m: make(map[*Stream]struct{}),
+		},
```
Kubuxu: same pattern of map initialization here.

swarm.go
```diff
@@ -4,187 +4,169 @@ package swarm
...
 // be opened and closed, while still using the same Chan for all
 // communication. The Chan sends/receives Messages, which note the
 // destination or source Peer.
-//
-// Uses peerstream.Swarm
 type Swarm struct {
-	swarm *ps.Swarm
+	// Close refcount. This allows us to fully wait for the swarm to be torn
+	// down before continuing.
+	refs sync.WaitGroup
```
Kubuxu: I am not sure if WG is right thing to use for this. If Wait was already called and Add is called after it will panic.

swarm_stream.go
```diff
@@ -1,69 +1,165 @@
...
+)
+
+// Stream is the stream type used by swarm. In general, you won't use this type
+// directly.
+type Stream struct {
+	stream smux.Stream
+	conn   *Conn
+
+	stateMu sync.Mutex
+	state   streamState
```
Kubuxu: do you want to apply the same anonymous type pattern here?

swarm.go
Stebalien: Good point, probably easier to read. A waitgroup *should* work as we only call this *after* preventing anything else from starting but yeah, a rwlock is definitely simpler.

dial_sync_test.go
Stebalien: import grouping (std, remote, local).

swarm.go
Stebalien: On further inspection, the code is currently *technically* correct because add only panics if the count reaches zero and then goes above zero). Really, we *want* that behaviour. However, the ref tracking code is still really hard to follow so I'll see what I can do.

Stebalien: I tried... there are now more comments. I can't see a nicer (while still efficient) way to do this.

dial_sync_test.go
Stebalien: Oh, wait, that's wrong. Should have been std, local, remote (fixed).

PR #70, opened 2018-06-07
feat: lower timeout for local address dials

magik6k:
Part of https://github.com/libp2p/go-libp2p/issues/1553
I didn't add ipv6 to the filters as nodes don't seem to announce the link-local addresses (or we filter them out elsewhere)

Tried this locally, seems to help a bit (limiter no longer fills with local addresses)

* Is there a better place for address matching logic?

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -22,9 +22,15 @@ import (
 )
 
 // DialTimeout is the maximum duration a Dial is allowed to take.
+// This includes the time spent waiting in dial limiter, between dialing the raw
+// network connection, protocol selection as well the handshake, if applicable.
+var DialTimeout = 60 * time.Second
+
+// DialTimeoutLocal is the maximum duration a Dial to local network address
+// is allowed to take.
 // This includes the time between dialing the raw network connection,
 // protocol selection as well the handshake, if applicable.
-var DialTimeout = 60 * time.Second
+var DialTimeoutLocal = 5 * time.Second
 
 var log = logging.Logger("swarm2")
 
--- a/addrs.go
+++ b/addrs.go
@@ -1,0 +1,35 @@
+package swarm
+
+import (
+	mafilter "github.com/libp2p/go-maddr-filter"
+	mamask "github.com/whyrusleeping/multiaddr-filter"
+)
+
+// http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
+var lowTimeoutFilters = mafilter.NewFilters()
+
+func init() {
+	for _, p := range []string{
+		"/ip4/10.0.0.0/ipcidr/8",
+		"/ip4/100.64.0.0/ipcidr/10",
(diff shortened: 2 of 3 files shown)


swarm.go
```diff
@@ -26,6 +26,11 @@ import (
 // protocol selection as well the handshake, if applicable.
 var DialTimeout = 60 * time.Second
 
+// DialTimeoutLocal is the maximum duration a Dial to local network address
+// is allowed to take. This only includes establishing the raw network
```
Stebalien: I'm pretty sure this dial timeout will also include negotiating the security protocol etc.

Stebalien:
Something like what you're doing here should work as an interum solution. As a long term solution, I'd like to more generally track expected RTTs.

I wonder if we could:

1. Have some kind of RTT tracking service.
2. Expose this service to transports. We can do this by feeding the service through with the context when dialing or by passing it into the transports when constructing them (as we do with the upgrader). Personally, I prefer the first option.

When dialing, services would:

1. Calculate timeouts based on expected RTTs reported by this service.
2. Report actual RTTs back to this service.

swarm.go
magik6k: oh, right

whyrusleeping:
@magik6k heres the tool i've been using to debug this (and generate those dial logs): https://github.com/libp2p/dht-utils/blob/master/query-test/main.go

(sorry no docs, you might have to `gx-go uw` in go-ipfs for it to build)

whyrusleeping:
While we're on this topic. Should we talk about just not dialing peers in private subnets that we arent also in?

I'm fine merging this change as is, but maybe the next thing to do would be that?

magik6k:
> While we're on this topic. Should we talk about just not dialing peers in private subnets that we arent also in?

I'd say that this should be done at the individual transport level, probably with some common helpers put to `go-addr-util`.

One problem is that in larger networks hosts tend to not know all lan routes. Simplest example would be `[internets] <- NAT <- 10.0.0.0/8 net <- NAT <- 162.168.0.0/24 net <- ipfs instance`. We'd now break connectivity to hosts in 10.0.0.0/8 network.

whyrusleeping:
> One problem is that in larger networks hosts tend to not know all lan routes. Simplest example would be `[internets] <- NAT <- 10.0.0.0/8 net <- NAT <- 162.168.0.0/24 net <- ipfs instance`. We'd now break connectivity to hosts in 10.0.0.0/8 network.

Yeah... good point.

whyrusleeping (approved):
Changes look good to me. Would like to give @bigs and @Stebalien (if he wants) a chance to review and 👍 too.

bigs (approved):
Looks great! Super clean & well orchestrated with the child `Context`

whyrusleeping:
alright, let's ship this :)

whyrusleeping:
Thanks @magik6k!

PR #71, opened 2018-06-08
make the dial timeout apply per dial attempt

Stebalien:
and switch to using the DialTimeout from go-libp2p-transport

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -221,9 +221,6 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
 		return c, nil
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, DialTimeout)
-	defer cancel()
-
 	logdial := lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
 
 	// ok, we have been charged to dial! let's do it.
@@ -259,6 +256,9 @@ func (s *Swarm) canDial(addr ma.Multiaddr) bool {
 
 // dial is the actual swarm's dial logic, gated by Dial.
 func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
+	ctx, cancel := context.WithTimeout(ctx, DialTimeout)
+	defer cancel()
+
 	var logdial = lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
 	if p == s.local {
 		log.Event(ctx, "swarmDialDoDialSelf", logdial)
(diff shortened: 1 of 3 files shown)


swarm_dial.go
```diff
@@ -221,9 +221,6 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
 		return c, nil
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, DialTimeout)
```
magik6k: I'd still keep this timeout around, just make it really big (like 10 min (10x dial timeout)).

Best way to handle this imo would be to have some way tell the context to only count down when we actually do the dialing, we could then set this limit to e.g. 5 'dial-minutes'

magik6k (commented):
Code LGTM, This should improve things, just not sure about removing the 'global' dial timeout.

magik6k:
It's trying to fix the issue described in https://github.com/libp2p/go-libp2p-swarm/issues/68#issuecomment-395501312, I'd keep the timeout and make it large

With https://github.com/libp2p/go-tcp-transport/pull/24 this shouldn't create any real problem though

whyrusleeping:
@Stebalien Yeah, i would reduce the per-dial timeout. Probably play it safe and start with 30 seconds, though I suspect numbers as low as 5 seconds work better (as long as its not hard for users to override this in a specific `ipfs swarm connect` call)

whyrusleeping:
@Stebalien I know, but it shouldnt be too hard to do by wiring down context hints. And that can be done in the future

Stebalien:
So, I've changed my mind on this (having a global peer dial timeout):

> I agree that having a timeout there is probably a good idea. However, as far as I can tell I added (well, moved) this timeout in the refactor so we'll have to think carefully about it.

The way it's currently implemented, it will cause *waiting* dials to timeout, even if they haven't done anything yet. This can lead to a cascading dial failure situation where dials pile up and then timeout while in the queue.

Ideally, we'd have a time and resource accounting system. However, any reasonable implementation would take a non-trivial amount of work to implement.

Stebalien:
> The way it's currently implemented, it will cause waiting dials to timeout, even if they haven't done anything yet. This can lead to a cascading dial failure situation where dials pile up and then timeout while in the queue.

Actually, the other way to think about this is that client that initiated the dial probably doesn't care about it after a period of time.

Instead of going with the resource accounting system, I'm going to change the overall multi-dial timeout to a DialPeer timeout and make it apply to each call to DialPeer.

---

@whyrusleeping 

> I know, but it shouldnt be too hard to do by wiring down context hints.

This is actually a bit tricky because we coalesce multiple dials into one.

Stebalien:
Closing in favor of #77.

And I'm trying to centralize discussion in #76.

PR #74, opened 2018-06-13
log an error when we have no transports configured

Stebalien:
Modify our two transport lookup functions to log loudly when there are no
transports registered.


bigs:
rebased and merging, definitely happy to have more logs :)

bigs:
just a note, the tests passed on travis but failed due to a timeout on jenkins. i ran locally (gxed and all) and it was fine, so i'm proceeding.

PR #80, opened 2018-06-26
remove redundant dial timeout

Stebalien:
This timeout is already set by the dial limiter.


Stebalien:
(hidden conflict with your PR that I missed)

PR #84, opened 2018-08-16
Issue #370 Check if empty peer ID

kishansagathiya:
On trying to open a new stream against an empty peer ID, we get an
error saying that failed to dial a peer.

This commit changes that to a more informative error message:
`empty peer ID`

Part of the fix for https://github.com/libp2p/go-libp2p/issues/370


Stebalien:
LGTM but blocked on bubbling a go-libp2p-peer update. I'll do that as soon as I get the chance and then I'll merge this.

Stebalien:
The tests pass locally, no real need to rebase and run them on CI.

Stebalien:
(and I've updated the go-libp2p-peer dep so this is good to go)

PR #95, opened 2019-01-23
avoid spawning goroutines for canceled dials

Stebalien:
This may speed up working through our dial queue a bit (but probably isn't the main issue).

### DIFF
--- a/limiter.go
+++ b/limiter.go
@@ -77,17 +77,26 @@ func newDialLimiterWithParams(df dialfunc, fdLimit, perPeerLimit int) *dialLimit
 func (dl *dialLimiter) freeFDToken() {
 	dl.fdConsuming--
 
-	if len(dl.waitingOnFd) > 0 {
+	for len(dl.waitingOnFd) > 0 {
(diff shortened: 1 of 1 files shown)


limiter.go
```diff
@@ -77,17 +77,26 @@ func newDialLimiterWithParams(df dialfunc, fdLimit, perPeerLimit int) *dialLimit
...
+
 		if len(dl.waitingOnFd) == 0 {
-			dl.waitingOnFd = nil // clear out memory
+			// clear out memory.
+			dl.waitingOnFd = nil
+		}
+
+		// Skip over canceled dials instead of queuing up a goroutine.
+		if next.cancelled() {
+			dl.freePeerToken(next)
```
raulk: Good catch, the peer token wasn't being freed in the case of a cancellation.

Stebalien: I don't think *that* was the issue. We'd (a) spawn an `executeDial` goroutine and then (b) free the token when we returned from it. *This* PR just avoids spawning goroutines that we don't need.

PR #96, opened 2019-01-23
dialer: handle dial cancel and/or completion before trying new addresses

Stebalien:
Otherwise, we can continue dialing even if we want to stop.

Note: The dials will all get aborted early so this probably isn't causing any significant performance issues.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -358,14 +358,34 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
 	defer s.limiter.clearAllPeerDials(p)
 
 	var active int
-	for {
+	for remoteAddrs != nil || active > 0 {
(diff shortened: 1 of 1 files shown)


swarm_dial.go
```diff
@@ -359,6 +359,33 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
...
+				if remoteAddrs == nil && active == 0 {
+					return nil, exitErr
+				}
+			} else if resp.Conn != nil {
+				return resp.Conn, nil
+			}
+
+			// We got a result, try again from the top.
+			continue
+		default:
```
Stebalien: This is a bunch of duplicated code. Not sure if there's a better way.

raulk: The context cancellation check is crucial. Draining responses first before starting more dials is an optimisation, but a welcome one.

PR #100, opened 2019-02-04
Fix logging race

anacrolix:
Fix race caused by log statement. Merge #99 to see its effect.

### DIFF
--- a/limiter.go
+++ b/limiter.go
@@ -203,10 +203,6 @@ func (dl *dialLimiter) clearAllPeerDials(p peer.ID) {
 // channel when finished. Once the response is sent it also releases all tokens
 // it held during the dial.
 func (dl *dialLimiter) executeDial(j *dialJob) {
-	dl.lk.Lock()
-	log.Debugf("[limiter] executing dial (dialfunc); peer: %s; addr: %s; FD consuming: %d; waiting: %d",
-		j.peer, j.addr, dl.fdConsuming, len(dl.waitingOnFd))
-	dl.lk.Unlock()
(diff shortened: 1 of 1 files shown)


limiter.go
```diff
@@ -203,8 +203,10 @@ func (dl *dialLimiter) clearAllPeerDials(p peer.ID) {
 // channel when finished. Once the response is sent it also releases all tokens
 // it held during the dial.
 func (dl *dialLimiter) executeDial(j *dialJob) {
+	dl.lk.Lock()
 	log.Debugf("[limiter] executing dial (dialfunc); peer: %s; addr: %s; FD consuming: %d; waiting: %d",
 		j.peer, j.addr, dl.fdConsuming, len(dl.waitingOnFd))
+	dl.lk.Unlock()
```
vyzo: what's the race here? `log` should be thread-safe.

raulk:
I’m ok with this race. It’s not worthwhile to sync for a log that does not require consistency and that will not be printed in 99% of cases (debug). Even if we sync’ed, it would not have a visible effect. Any way to quieten the race checker? Reopen if you think this is a mistake.

Kubuxu:
Unfortunately, there is no way to quiet the racechecker. Maybe we should just remove the log, the dial will be logged esewhere.

raulk:
@Kubuxu I'm not one to remove logs, but it's justified in this case because this statement is actually superfluous. The original one is here: https://github.com/libp2p/go-libp2p-swarm/blob/master/limiter.go#L162, and we trust Go enough to launch the goroutine. So this statement adds no value.

Issue #69, opened 2018-06-07
Change the default error returned by `dialAddrs`


Stebalien:
"default failure" is really uninformative. I believe the error case here is *actually* "no addresses found for peer X".

anacrolix:
Fixed by #109.

PR #101, opened 2019-02-04
Improve error messages when remote addresses are bad

anacrolix:
Improves the error message when there are no addresses because they're all filtered out, or none are given. Also the dial logic is simplified. Timing, and pseudo-random channel operations make optimistic selects unworthwhile.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -290,6 +290,9 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		that we previously had (halting a dial when we run out of addrs)
 	*/
 	goodAddrs := s.filterKnownUndialables(s.peers.Addrs(p))
+	if len(goodAddrs) == 0 {
+		return nil, errors.New("no good addresses")
+	}
 	goodAddrsChan := make(chan ma.Multiaddr, len(goodAddrs))
 	for _, a := range goodAddrs {
 		goodAddrsChan <- a
@@ -343,71 +346,46 @@ func (s *Swarm) filterKnownUndialables(addrs []ma.Multiaddr) []ma.Multiaddr {
 	)
 }
 
-func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.Multiaddr) (transport.Conn, error) {
+func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.Multiaddr) (conn transport.Conn, err error) {
 	log.Debugf("%s swarm dialing %s", s.local, p)
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel() // cancel work when we exit func
 
-	// use a single response type instead of errs and conns, reduces complexity *a ton*
-	respch := make(chan dialResult)
+	dialResults := make(chan dialResult)
 
-	defaultDialFail := fmt.Errorf("failed to dial %s (default failure)", p)
-	exitErr := defaultDialFail
+	var lastDialErr error
 
 	defer s.limiter.clearAllPeerDials(p)
 
 	var active int
 	for remoteAddrs != nil || active > 0 {
-		// Check for context cancellations and/or responses first.
-		select {
-		case <-ctx.Done():
-			if exitErr == defaultDialFail {
-				exitErr = ctx.Err()
-			}
-			return nil, exitErr
-		case resp := <-respch:
-			active--
-			if resp.Err != nil {
-				log.Infof("got error on dial to %s: %s", resp.Addr, resp.Err)
-				// Errors are normal, lots of dials will fail
-				exitErr = resp.Err
-			} else if resp.Conn != nil {
-				return resp.Conn, nil
-			}
-
-			// We got a result, try again from the top.
-			continue
-		default:
-		}
-
-		// Now, attempt to dial.
 		select {
 		case addr, ok := <-remoteAddrs:
 			if !ok {
 				remoteAddrs = nil
 				continue
 			}
-
-			s.limitedDial(ctx, p, addr, respch)
+			s.limitedDial(ctx, p, addr, dialResults)
 			active++
 		case <-ctx.Done():
-			if exitErr == defaultDialFail {
-				exitErr = ctx.Err()
+			if lastDialErr != nil {
+				return nil, lastDialErr
+			} else {
+				return nil, ctx.Err()
 			}
-			return nil, exitErr
-		case resp := <-respch:
+		case resp := <-dialResults:
 			active--
 			if resp.Err != nil {
 				log.Infof("got error on dial to %s: %s", resp.Addr, resp.Err)
 				// Errors are normal, lots of dials will fail
-				exitErr = resp.Err
-			} else if resp.Conn != nil {
+				lastDialErr = resp.Err
+			} else {
 				return resp.Conn, nil
 			}
 		}
 	}
-	return nil, exitErr
+	return nil, errors.New("no remote addresses")
 }
 
 // limitedDial will start a dial to the given peer when
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -385,7 +385,11 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
 			}
 		}
 	}
-	return nil, errors.New("no remote addresses")
+	if lastDialErr != nil {
+		return nil, lastDialErr
+	} else {
+		return nil, errors.New("no remote addresses")
+	}
 }
 
 // limitedDial will start a dial to the given peer when
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -364,16 +364,14 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
 			// no dials succeed.
 			lastDialErr = dr.Err
 			return false
-		} else {
-			return true
 		}
+		return true
 	}
 	ctxDoneRetErr := func() error {
 		if lastDialErr != nil {
 			return lastDialErr
-		} else {
-			return ctx.Err()
 		}
+		return ctx.Err()
 	}
 	for remoteAddrs != nil || activeDials > 0 {
 		// Try to proceed on context completion or a dial result before dialing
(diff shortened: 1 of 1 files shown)


swarm_dial.go
```diff
@@ -343,71 +346,50 @@ func (s *Swarm) filterKnownUndialables(addrs []ma.Multiaddr) []ma.Multiaddr {
...
-				exitErr = resp.Err
-			} else if resp.Conn != nil {
-				return resp.Conn, nil
-			}
-
-			// We got a result, try again from the top.
-			continue
-		default:
-		}
-
```
vyzo: why was this code removed?

swarm_dial.go
```diff
@@ -290,6 +290,9 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		that we previously had (halting a dial when we run out of addrs)
 	*/
 	goodAddrs := s.filterKnownUndialables(s.peers.Addrs(p))
+	if len(goodAddrs) == 0 {
+		return nil, errors.New("no good addresses")
```
Kubuxu: IMO, use exported error value instead of inline errors.

swarm_dial.go
```diff
@@ -343,71 +346,50 @@ func (s *Swarm) filterKnownUndialables(addrs []ma.Multiaddr) []ma.Multiaddr {
 	)
 }
 
-func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.Multiaddr) (transport.Conn, error) {
+func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.Multiaddr) (conn transport.Conn, err error) {
```
Kubuxu: Are the named return values needed?

swarm_dial.go
anacrolix: Sure, if they have value to consumers. Here I'm simply providing a better textual error than was previously available. I don't want to commit prematurely to any meaning.

swarm_dial.go
anacrolix: Because it makes assumptions about timing that aren't worthwhile (instead there's just a lot of noisy code duplication). If it's a strict requirement that (which it isn't) that dial results be consumed before dials, the correct concurrency mechanisms should be used.

swarm_dial.go
anacrolix: Thanks, no.

swarm_dial.go
Kubuxu: I think we were doing this because dial might have finished before we started the next one. There is no guarantee of order you will receive messages in the select block.

You might first get all messages from `remoteAddrs` before consuming any from `respch` or context.

This is a use of concurrency mechanism.

Kubuxu: Which means, even if you've got a valid connection or context cancellation, without this additional loop we might continue dialling, which is just wasteful.

vyzo: @anacrolix please reinstate this code.

swarm_dial.go
raulk: Allowing consumers to distinguish when a dial failed is valuable. Precisely by exporting this error you won't be committing to anything because you can change the text anytime without breaking consumers. If you don't export it, you set yourself up for users doing textual comparisons on the error message if they wan't to distinguish this situation.

swarm_dial.go
raulk: In fact, this was added recently precisely to drain finished dials before starting new ones, thus making a better use of resources: https://github.com/libp2p/go-libp2p-swarm/pull/96

swarm_dial.go
```diff
@@ -349,65 +352,44 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
...
-		case resp := <-respch:
+		case resp := <-dialResults:
 			active--
 			if resp.Err != nil {
 				log.Infof("got error on dial to %s: %s", resp.Addr, resp.Err)
 				// Errors are normal, lots of dials will fail
-				exitErr = resp.Err
-			} else if resp.Conn != nil {
+				lastDialErr = resp.Err
+			} else {
```
raulk: @Stebalien any clue why we had an else-if here instead of an else?

swarm_dial.go
```diff
@@ -349,65 +352,44 @@ func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.
...
 				return resp.Conn, nil
 			}
 		}
 	}
-	return nil, exitErr
+	if lastDialErr != nil {
+		return nil, lastDialErr
+	} else {
+		return nil, errors.New("no remote addresses")
```
raulk: This is a validation error and it should be caught at the beginning for clarity. Also, let's make it an exported error.

swarm_dial.go
anacrolix: That's fair enough, but is it even an actionable error?

swarm_dial.go
anacrolix: Did it actually have a measurable effect? As I mentioned above, there are better alternative concurrency mechanisms that would have a more reliable impact. The "prioritized" select changed in #96 is optimistic at best. There's still a race between the first and second select, and any benefit relies on a whole bunch of assumptions about timings.

swarm_dial.go
anacrolix: It's usually a defence against a bad interface implementation. I've had to do similar things in the past. I've opted not to persist it here.

swarm_dial.go
anacrolix: Actually I think it's better to put it at the end. In this instance it's flexing a constraint of this method: If there are any remote addresses and none of them connect, then by we should have a dial error. Putting it at the end here reduces the special conditions on this method.

swarm_dial.go
Stebalien: There's always going to be a race here and there's nothing we can do about it, regardless of the synchronization mechanism. However, the extra select means that we'll start at most one unnecessary dial. Without it, we could end up draining `remoteAddrs` before we stop dialing.

swarm_dial.go
Stebalien: No idea, that's been there for ages. I'm pretty sure we always return one or the other.

anacrolix:
Fixes #69.

swarm_dial.go
anacrolix: I ran the code with panics in place to detect when dials had actually completed and we pushed another dial job. It was exceptionally rare, and even then sits on the dial queue before being cancelled out. I've rehashed the entire function to make the priorities explicity without duplicating the dial response processing, I don't know if it's any better. Please take a look.

anacrolix:
I've refactored this back to the channel model, with the optimistic select. I need the more specific error messages as "default failure" is getting really old really quick and making debugging a pain.

swarm_dial.go
raulk: Yes, it is.

swarm_dial.go
```diff
@@ -345,69 +348,70 @@ func (s *Swarm) filterKnownUndialables(addrs []ma.Multiaddr) []ma.Multiaddr {
...
-				log.Infof("got error on dial to %s: %s", resp.Addr, resp.Err)
-				// Errors are normal, lots of dials will fail
-				exitErr = resp.Err
-			} else if resp.Conn != nil {
-				return resp.Conn, nil
+			return nil, ctxDoneRetErr()
+		case dr := <-dialResults:
+			if onDialResult(dr) {
+				return dr.Conn, nil
+			} else {
```
raulk: This `else` is superfluous.

swarm_dial.go
raulk: Superfluous `else`.

raulk:
I suggest sticking to introducing an **exported error** for when we have no remote addresses to dial.

anacrolix:
I'm on board with that error now, I'll add it.

anacrolix:
@raulk This error, if exposed, belongs in go-libp2p-net, as a part of the Dialer and Network interfaces there.

swarm_dial.go
anacrolix: @raulk How? You try to dial a peer and can't because you have no addresses for them. You don't then add them and try again, I don't think anybody does that. If you get a dial error because you had no good addresses, you should have already added all the addresses you can think of.

swarm_dial.go
anacrolix: Because this is a pattern match, and not a guard, like you would normally apply this rule in Go. It's a stylistic choice I've borrowed from Haskell. Neither boolean value is exceptional here, I'm only acting on the control flow indication provided.

swarm_dial.go
anacrolix: Sure, this one is a guard.

swarm_dial.go
anacrolix: I removed it anyway, it looks nicer.

anacrolix:
> @anacrolix I don't like this mixture of functional and procedural (and the added indirection) just to shortcut 2-3 lines of code. To me, the value prop is not there, especially given that repetition in Go is idiomatic. I much rather read this linearly than seesawing up and down.

It's explicit about identical behaviour on different sets of communication clauses in the select. I think it's worth doing, especially with the extensive behaviour and comments in `onDialResult`. `ctxDoneRetErr` is there for symmetry. I don't want to bike-shed this part.

anacrolix:
This now requires https://github.com/libp2p/go-libp2p-net/pull/38 to be merged.

raulk:
@anacrolix 

> I don't want to bike-shed this part.

Isn't 90% of this PR a bike-shed of `master`?

First, would you mind making this PR stick to its title?

Second, you're proposing to introduce a **personal** stylistic preference on code that has been vetted by the community at large. As I explained above, IMO your change introduces extra bookkeeping variables, and makes it unreadable linearly. This is a critical part of the system where explicitness is appreciated. Optimistic selects are used largely in the codebase in their denormalised form. I also believe that, in Go, repetitiveness is OK and idiomatic. Moreover, the resulting block has no substantial reduction in LoC either, so by all accounts I really don't understand your insistence here.

To move forward, I propose to stop beating a dead horse. Can you close this PR and open a new one, just with the changes announced in the title, which are pretty non-controversial?

If you _really_ want to push the closure stuff, open a separate PR just for that and try to get others to chime in; you already know where I stand ;-)

PR #104, opened 2019-02-15
Trivial context and comments

anacrolix:
I've noticed a lot of locks around to ensure notification ordering. I think it's much superior to do synchronous notifications and allow consumers to decide for themselves how to handle concurrency. I've tested this significantly with the DHT, where it also provides for a lot of simplification. There is a fair bit of notification flow where go routines are spawned at each handler step, as an example.

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -121,8 +121,8 @@ func (s *Swarm) teardown() error {
 	s.conns.m = nil
 	s.conns.Unlock()
 
-	// Lots of go routines but we might as well do this in parallel. We want
-	// to shut down as fast as possible.
+	// Lots of goroutines but we might as well do this in parallel. We want to shut down as fast as
+	// possible.
 
 	for l := range listeners {
 		go func(l transport.Listener) {
@@ -148,8 +148,8 @@ func (s *Swarm) teardown() error {
 	return nil
 }
 
-// AddAddrFilter adds a multiaddr filter to the set of filters the swarm will
-// use to determine which addresses not to dial to.
+// AddAddrFilter adds a multiaddr filter to the set of filters the swarm will use to determine which
+// addresses not to dial to.
 func (s *Swarm) AddAddrFilter(f string) error {
 	m, err := mafilter.NewMask(f)
 	if err != nil {
--- a/swarm.go
+++ b/swarm.go
@@ -208,9 +208,6 @@ func (s *Swarm) addConn(tc transport.Conn, dir inet.Direction) (*Conn, error) {
 	// * The other will be decremented when Conn.start exits.
 	s.refs.Add(2)
 
-	// Take the notification lock before releasing the conns lock to block
-	// Disconnect notifications until after the Connect notifications done.
-	c.notifyLk.Lock()
 	s.conns.Unlock()
 
 	// We have a connection now. Cancel all other in-progress dials.
@@ -220,7 +217,6 @@ func (s *Swarm) addConn(tc transport.Conn, dir inet.Direction) (*Conn, error) {
 	s.notifyAll(func(f inet.Notifiee) {
 		f.Connected(s, c)
 	})
-	c.notifyLk.Unlock()
 
 	c.start()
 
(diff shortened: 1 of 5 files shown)


swarm.go
```diff
@@ -208,9 +208,6 @@ func (s *Swarm) addConn(tc transport.Conn, dir inet.Direction) (*Conn, error) {
 	// * The other will be decremented when Conn.start exits.
 	s.refs.Add(2)
 
-	// Take the notification lock before releasing the conns lock to block
-	// Disconnect notifications until after the Connect notifications done.
-	c.notifyLk.Lock()
```
Stebalien: As noted in the comment, this will allow disconnect events to fire (concurrently) before connect events. This is an important invariant to maintain and has caused us no end of grief in the past.

swarm_conn.go
```diff
@@ -65,17 +65,10 @@ func (c *Conn) doClose() {
...
-		// prevents us from issuing close notifications before finishing the open notifications
-		c.notifyLk.Lock()
-		defer c.notifyLk.Unlock()
-
-		c.swarm.notifyAll(func(f inet.Notifiee) {
-			f.Disconnected(c.swarm, c)
-		})
-		c.swarm.refs.Done() // taken in Swarm.addConn
-	}()
+	c.swarm.notifyAll(func(f inet.Notifiee) {
```
Stebalien: This is _currently_ asynchronous for the reason specified in the comment.

swarm.go
```diff
@@ -438,18 +434,10 @@ func (s *Swarm) Backoff() *DialBackoff {
...
 	for f := range s.notifs.m {
-		go func(f inet.Notifiee) {
-			defer wg.Done()
-			notify(f)
-		}(f)
+		notify(f)
 	}
-
-	wg.Wait()
 	s.notifs.RUnlock()
```
Stebalien: This is probably fine but we should profile this in an application (e.g., go-ipfs) as we've had a _ton_ of issues around notifications. Given that, I'm really not sure this change is worth it. Short-lived goroutines aren't _that_ expensive.

Kubuxu: Recently I saw bitswap lock up about 400 notifications, it will be better after the current rewrite.

---

Also I think if we leave the goroutine code there, we can move the `RUnlock()` above `wg.Wait()`

Stebalien: How recently and are you saying that doing these in series (this patch) will make it better?

Kubuxu: Two days ago but it was on a 0.4.18.
No it won't, this node would have probably died if not for it. Bitswap should handle those notifications with priority (loop over select with default). I think it will be better now after the rewrite.

anacrolix: Spawning unbounded goroutines with the assumption that "they'll get processed" is not a good idea. The propagation of notifications needs to be overhauled (better yet, reduced, there is an awful lot of abstraction inversion floating around).

swarm.go
anacrolix: Are there unit tests for this?

swarm.go
anacrolix: Do you intend to create an issue for this? Is there one in IPFS we can observe?

anacrolix:
I've removed the notification stuff, and just left the trivial stuff in.

swarm.go
Stebalien: Not that I know of.

swarm.go
Kubuxu: As there was go-bitswap rewrite I will wait for it to reappear in future.

Kubuxu:
Another option might be to require the consumer of the API to provide a channel where notifications will be sent to. This channel has to have enough capacity to not block notification goroutine, otherwise, notification will be dropped.

It changes the semantics around the notifications and might be problematic.

anacrolix:
@Kubuxu I don't understand why the notifications must be asynchronous. Let the consumer control the concurrency situation.

raulk:
@Kubuxu that approach sounds reasonable, but I don't trust the Notifee to _not_ close the channel before time. If they do, the system will panic and it gets ugly. And a Notifee can be anything, i.e. user code.

Kubuxu:
> I don't understand why the notifications must be asynchronous. Let the consumer control the concurrency situation.

because one misbehaving consuming system affects the performance of the whole application

> I don't trust the Notifee to _not_ close the channel before time

This is API contract you have to make with your users with clean docs. There are multiple ways to achieve similar results when users miss uses an API (for example, taking a pointer to a variable and not copying it).

Not blocking this PR, just pointing out a possible alternative.

PR #113, opened 2019-03-08
Differentiate no addresses error from no good addresses

anacrolix:
Address filtering is applied before considering returning an error. This adds an extra error if there are no addresses at all to differentiate this. I found this useful as sometimes addresses are filtered for having trailing /p2p components, which was confusing. A later PR may improve that too.

The error check is changed to determine if the backoff could be part of context completion, as opposed to returned by an arbitrary downstream context (but not part of the current context), which I believe to the real intention.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -248,7 +248,7 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
 			log.Debugf("ignoring dial error because we have a connection: %s", err)
 			return conn, nil
 		}
-		if ctx.Err() == nil {
+		if err != context.Canceled {
 			log.Event(ctx, "swarmDialBackoffAdd", logdial)
 			s.backf.AddBackoff(p) // let others know to backoff
 		}
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -84,7 +84,7 @@ const DefaultPerPeerRateLimit = 8
 
 // DialBackoff is a type for tracking peer dial backoffs.
 //
-// * It's safe to use it's zero value.
+// * It's safe to use its zero value.
 // * It's thread-safe.
 // * It's *not* safe to move this type after using.
 type DialBackoff struct {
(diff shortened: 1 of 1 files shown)


swarm_dial.go
```diff
@@ -289,7 +289,11 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		the improved rate limiter, while maintaining the outward behaviour
 		that we previously had (halting a dial when we run out of addrs)
 	*/
-	goodAddrs := s.filterKnownUndialables(s.peers.Addrs(p))
+	peerAddrs := s.peers.Addrs(p)
+	if len(peerAddrs) == 0 {
+		return nil, errors.New("no addresses")
```
raulk: Exported error, please.

swarm_dial.go
```diff
@@ -248,7 +248,7 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
 			log.Debugf("ignoring dial error because we have a connection: %s", err)
 			return conn, nil
 		}
-		if err != context.Canceled {
+		if ctx.Err() == nil {
```
raulk: Previous condition was correct; we don't want to add a backoff if the user cancelled the context.

swarm_dial.go
raulk: Appreciate if you can export the "no good addresses" error below too.

anacrolix: I just don't see it as an actionable error, and not worth littering the net interface for.

swarm_dial.go
anacrolix: That's precisely what the new code does. The old code checks for only one of the two (currently) possible context completion errors, that may not have been triggered by the user's context.

raulk: That’s because we *want* to add the backoff if the deadline is exceeded (ie. timeout), but not if the user manually cancels the context.

swarm_dial.go
raulk: Precisely due to the lack of an exported error here, other modules check the peerstore to determine if we have addresses for a peer. That can lead to false positives where we have addresses but none are diallable. Not worth to fix that now, but I did want to colour my reasoning.

anacrolix: Maybe we can address this later, when the need arises.

swarm_dial.go
anacrolix: I've removed this change.

anacrolix:
I'll squash this one when it's approved.

raulk:
@anacrolix Oh well, too late for squashing.

PR #115, opened 2019-03-26
return all dial errors if dial has failed

reinerRubin:
Have tried to solve this issue [1] ("Print dial status messages to stderr"). I have not changed "Info" level because it seems that dial errors happen all the time (see comments around). Instead dial collects dial errors and if all attempts fail returns combined error. Printing them would be client concern.  

But, idk, about context. Maybe in case of total dial fail it would be better just printing collected errors to stderr like it was proposed in the ticket.

Also the test seems useless, but at least it allows play around with the problem.

1 - https://github.com/ipfs/go-ipfs/issues/4355

### DIFF
--- a/dial_test.go
+++ b/dial_test.go
@@ -2,7 +2,9 @@ package swarm_test
 
 import (
 	"context"
+	"fmt"
 	"net"
+	"regexp"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


swarm_dial.go
```diff
@@ -457,3 +460,16 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
 	// success! we got one!
 	return connC, nil
 }
+
+func combineAddressDialErrors(defaultError error, dialErrors ...error) error {
```
Stebalien: FYI, we have some other packages that use https://github.com/hashicorp/go-multierror.

swarm_dial.go
```diff
@@ -65,20 +66,20 @@ const DefaultPerPeerRateLimit = 8
...
-//  		continue
-//  	}
-//  	defer dialsync.Unlock(p)
-//  	c, err := actuallyDial(p)
-//  	if err != nil {
-//  		dialbackoff.AddBackoff(p)
-//  		continue
-//  	}
-//  	dialbackoff.Clear(p)
+//	if ok, wait := dialsync.Lock(p); !ok {
```
Stebalien: ?

swarm_dial.go
reinerRubin: Cool, I have tried to find something like this with grep but failed (only custom NewMultiError at core/corerepo/gc.go). Will change here to hashicorp/go-multierror.

swarm_dial.go
reinerRubin: My bad. Editor has strict extra spaces policy. Will also fix.

swarm_dial.go
Stebalien: Yeah, I brought this up because I'd like to switch over to some consistent multi-error package.

reinerRubin:
Ok, moved to "hashicorp/go-multierror" and also removed the trashy diff.

reinerRubin:
Can I make a test that calls directly "func (s *Swarm) dialAddrs" in the main swarm package? Because some code paths are not very testable. 

Is it ok that go.mod and gx (package.json) links to different repos ("github.com/hashicorp/go-multierror" and "gxed")?

anacrolix:
What value is there in collecting more than one error? Are they actually actionable? I believe the idea of multi-errors is a bit of an anti-pattern. If you wish to perform some action when a specific dial fails, you should be performing the dialing routine yourself so you have more information and control at hand, or expose more appropriate tracing/logging inside the existing dialing routine for observability (noting that you don't actually act on this immediately, apart from manually noting the reason for failures, and maybe correcting your network situation OOB).

Reading https://github.com/ipfs/go-ipfs/issues/4355 I think better logging is what's called for.

Stebalien:
Logs won't help much here. We need to expose these dial errors to the user.

Really, the _ideal_ way to do this is to feed some kind of dial status channel through with a context so the swarm can return information like "dialing peer X on address Y" and "failed to dial address Y: error". However, returning all the relevant errors instead of just one of them is _much_ better than the current code. At the moment, the user might be told something useless like "no route to host" when that's just one error among many other more relevant errors (e.g., "expected peer Y, found peer X").

raulk:
That’s similar to the idea proposed here: https://github.com/libp2p/go-libp2p-kad-dht/issues/300. Also ties in with an eventual eventbus at the host level.

I agree multierror is appropriate now.

reinerRubin:
Ok, so what about the testing policy? Can I add direct test to the package method (dialAddrs). Because idk how to cover diff with only public methods.

And is it ok that go.mod and gx (package.json) links to different repos ("github.com/hashicorp/go-multierror" and "gxed")?

Stebalien:
> Ok, so what is about the testing policy? Can I add direct test to the package method (dialAddrs). Because idk how to cover diff with only public methods.

You should be able to add two addresses for an undailable peer using the peerstore (`swarm.Peerstore().AddAddrs(...)`). You can then dial the peer normally and check the error.

> And is it ok that go.mod and gx (package.json) links to different repos ("github.com/hashicorp/go-multierror" and "gxed")?

Yes. The package.json file is deprecated.

reinerRubin:
I have added silent addresses and the tests are fine. But I do not know how to cover "context.Cancel()" and empty addresses cases only with public methods. It must be done to meet the "diff cover" requirement (codecov/patch — 60% of diff hit (target 77.43%)).

dial_test.go
```diff
@@ -480,3 +480,30 @@ func TestDialBackoffClears(t *testing.T) {
...
+			silentPeerAddress,
+			pstore.PermanentAddrTTL)
+	}
+
+	_, err := testedSwarm.DialPeer(ctx, targetSwarm.LocalPeer())
+	if err == nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("correctly get a combined error: %s", err)
```
Stebalien: This should test that we're getting the _right_ errors.

reinerRubin: I thought about it. But was not sure about check. The only way I see is to check err.Error() content itself. Is it ok?

Stebalien: Sounds good. Unfortunately, that's usually how error checking works in go. Ideally, there'd be two different errors to test for.

reinerRubin: I have added the additional checks for Error() content.

Stebalien:
Thanks!

PR #116, opened 2019-04-10
Add context option to disable dialing when opening a new stream

vyzo:
This is essential for (non active) relays who under no circumstances want to initiate a dial to a peer.

It was necessitated by the discovery of a large number of goroutines doing recursive dialing in a relay.

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -37,6 +37,18 @@ var ErrSwarmClosed = errors.New("swarm closed")
 // transport is misbehaving.
 var ErrAddrFiltered = errors.New("address filtered")
 
+// ErrNoConn is returned when attempting to open a stream to a peer with the NoDial
+// option and no usable connection is available.
+var ErrNoConn = errors.New("no usable connection to peer")
+
+// ContextOption is the type of context options understood by the swarm
+type ContextOption string
+
+// NoDial is a context option that instructs the swarm to not attempt a new
+// dial when opening a stream. The value of the key should be a string indicating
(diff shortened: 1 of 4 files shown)


swarm.go
```diff
@@ -37,6 +37,18 @@ var ErrSwarmClosed = errors.New("swarm closed")
 // transport is misbehaving.
 var ErrAddrFiltered = errors.New("address filtered")
 
+// ErrNoConn is returned when attempting to open a stream to a peer with the NoDial
+// option and no usable connection is available.
+var ErrNoConn = errors.New("no usable connection to peer")
+
+// ContextOption is the type of context options understood by the swarm
+type ContextOption string
```
Stebalien: Nit: This should probably be a newtype of `struct{}`.

Also, we should probably put this in go-libp2p-net. Now that we're using go modules, this should be _easy_ (:tada:)

vyzo: They need to be comparable, whatever that implies. Also the example in godoc uses a string alias type, so I went with it.

vyzo: I'll update go-libp2p-net to provide the option.

vyzo: Moved to `go-libp2p-net` in https://github.com/libp2p/go-libp2p-net/pull/42

vyzo: Updated to use the nodial option in go-libp2p-net.

vyzo:
@stebalien LGTM'ed on mobile; merging.

PR #118, opened 2019-04-17
handle relay addr dialing intelligently

vyzo:
We have a problem with dialing relay addrs: we dial all relays in the address set simultianeously and end up being connected to all of them.
This unnecessary increases the load in relays, and we end up initiating a bunch of parallel dials to all relay addrs. Couple that with addrsplosion, and we make a mess trying to dial relays.

This patch implements a dial scheduler that schedules relay addrs accordingly:
- the schedules non-relay dials immediately, without any delay
- the relay address sets are split to connected and not connected
  - connected relay addrs are aggregated to `/p2p/QmRelay/p2p-circuit` and issued in order, with a 500ms delay
 - non-connected relay addrs are issued in groups by relay, with a 750ms delay.

TBD: tests.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"sync"
 	"time"
 
@@ -16,6 +17,7 @@ import (
 	peer "github.com/libp2p/go-libp2p-peer"
 	transport "github.com/libp2p/go-libp2p-transport"
 	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
 )
 
 // Diagram of dial sync:
@@ -57,6 +59,10 @@ const ConcurrentFdDials = 160
 // DefaultPerPeerRateLimit is the number of concurrent outbound dials to make
 // per peer
 const DefaultPerPeerRateLimit = 8
+
+// this is the protocol code for relay. we can't import go-libp2p-circuit b/c
+// of circular dependency.
+const P_CIRCUIT = 290
 
 // dialbackoff is a struct used to avoid over-dialing the same, dead peers.
 // Whenever we totally time out on a peer (all three attempts), we add them
@@ -283,14 +289,6 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		log.Debug("Dial not given PrivateKey, so WILL NOT SECURE conn.")
 	}
 
-	//////
-	/*
-		This slice-to-chan code is temporary, the peerstore can currently provide
-		a channel as an interface for receiving addresses, but more thought
-		needs to be put into the execution. For now, this allows us to use
-		the improved rate limiter, while maintaining the outward behaviour
-		that we previously had (halting a dial when we run out of addrs)
-	*/
 	peerAddrs := s.peers.Addrs(p)
 	if len(peerAddrs) == 0 {
 		return nil, errors.New("no addresses")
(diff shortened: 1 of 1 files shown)


raulk:
Implementing the above feels like we'd be poor-man mimicking dialer v2. I think I have the bandwidth this week to take it over the finish line.

vyzo:
I totally agree but that's a little too hard with the current dialer, it will need a whole bunch of refactorings... which is what we are doing in dialer v2.
This is an interim solution after all with v2 in the pipeline.

raulk:
How urgently do we need this? Can it wait 1 week for a cleaner solution (dialer v2)?

vyzo:
We can wait a week, there is no imminent ipfs release. But I fear v2 won't be ready by then either :p

vyzo:
And also, I don't want to rush you on this. We can live with the patch for a while.

raulk:
Yeah, it's had a rocky history, due to my focus being all over the place :-( But it's clear we have increased pressure and impetus on it now, so I can commit to delivering it within a week.

raulk:
^^ may be reckless to give a concrete 1 week horizon. Let's say 2.

vyzo:
ok, this patch can stay in the queue for a week. If it turns out v2 is not gonna be ready for a while yet longer, we can merge and go ahead.

Also, please take into account the new logic about relay dialing: We want to prioritize connected relays and aggregate their addrs.

upperwal:
Hey @raulk 

If it ease out some of your burden, you can commit the intermediate code to some branch along with the TODO's you think needs to be done for the master branch. Contributors can help you achieve those tasks. I am interested to work on the dialer v2 logic and would be happy to contribute.

raulk:
@upperwal thanks for the offer, very much appreciated. Given the nature of the change, it’s best if executed by a single individual. I suggest kicking off a discussion in the forums for the community to propose projects where help is needed: https://discuss.libp2p.io

upperwal:
@raulk 👍

Stebalien:
@vyzo _should_ I review this?

vyzo:
@Stebalien I would say take a look for the scheduling logic, as we want it to carry over to dialer v2.

swarm_dial.go
```diff
@@ -352,6 +349,157 @@ func (s *Swarm) filterKnownUndialables(addrs []ma.Multiaddr) []ma.Multiaddr {
...
+		if err != nil {
+			immediateAddrs = append(immediateAddrs, a)
+			continue
+		}
+
+		relayS, err := a.ValueForProtocol(ma.P_P2P)
+		if err != nil {
+			// uspecfic relay addr; push it to immediate -- note that these addrs should not be
+			// advertised any more, but we don't want to break older nodes or tests
+			immediateAddrs = append(immediateAddrs, a)
```
Stebalien: Shouldn't we just drop this?

vyzo: Maybe, I really don't know what to do with these!

vyzo: Maybe we should schedule them for last, after all specific relay addrs have been scheduled.

Stebalien: Do we really gain anything? This literally means: dial me through random relays.

vyzo: Probably not. 
Only concern is that we do "support" users requesting dials with `ipfs swarm connect /p2p-circuit/ipfs/QmTarget`, and we may not want to completely break that.

Stebalien: Gah. What if we just hack that into a higher layer? Or just drop it? Nobody should be relying on that in production and I don't feel all that bad about breaking demos.

vyzo: Maybe we should drop it -- but we might want to disable circuit.OptDiscovery in ipfs as well.
We can substitute it with `/ipfs/QmRelay/p2p-circuit/ipfs/QmTarget` anyway.

Issue #119, opened 2019-04-21
Leaking dial errors


Stebalien:
I'm seeing a massive memory leak leaking errors returned by doDial. Not sure _where_ they're getting stashed, maybe in the dialer somewhere?

vyzo:
Maybe introduced by #115? We do aggregate all errors now.

vyzo:
Is there any more detail?

vyzo:
Also are there any stuck goroutines precluding the dial from finishing and accumulating all these errors?
I don't see any place where the errors could be stashed, other than the multierror contraption.

vyzo:
Maybe stuck in a channel though.

raulk:
Yeah, was thinking the same as @vyzo. Could you try with the commit previous to that PR being merged? Not sure if it’s a leak (us unintentionally retaining state) or just a side effect of accumulating all errors from the dial process (especially in the face of addrexplosion).

vyzo:
Can you also check for signs of https://github.com/libp2p/go-libp2p-quic-transport/issues/57? We do happen to have an open bug with stuck dials.

Stebalien:
> Maybe introduced by #115? We do aggregate all errors now.

This has almost certainly made it worse. However, we shouldn't be _storing_ these errors anywhere.

> Can you also check for signs of libp2p/go-libp2p-quic-transport#57? We do happen to have an open bug with stuck dials.

It's probably not that. This error is allocated _after_ all the transport dials finish so a stuck dial would prevent it.

raulk:
It could be the event logging of errors -- the weight of which is now augmented by multierrors. @Stebalien can you generate annotated source for these allocs?

Stebalien:
I don't have the traces but the buildup came from https://github.com/libp2p/go-libp2p-swarm/blob/94a49f124ed974ce57b9ed22c96ddba19a63b077/swarm_dial.go#L259.

PR #120, opened 2019-04-24
dial: limit error size

Stebalien:
Instead of storing _every_ error, store at most 32 errors (plus a "too many errors" error).

Helps address https://github.com/libp2p/go-libp2p-swarm/issues/119

Note: https://github.com/libp2p/go-libp2p-kad-dht/pull/328 actually _fixes_ the issue but this PR helps reduce the impact of similar issues.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -43,6 +43,9 @@ var (
 	// ErrNoTransport is returned when we don't know a transport for the
 	// given multiaddr.
 	ErrNoTransport = errors.New("no transport for protocol")
+
+	// ErrTooManyErrors is returned as the final error when we encounter too many errors when dialing a peer.
+	ErrTooManyErrors = errors.New("too many errors")
 )
 
 // DialAttempts governs how many times a goroutine will try to dial a given peer.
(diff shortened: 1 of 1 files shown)


swarm_dial.go
```diff
@@ -58,6 +61,9 @@ const ConcurrentFdDials = 160
 // per peer
 const DefaultPerPeerRateLimit = 8
 
+// maxDialErrors is the maximum number of dial errors we record
+const maxDialErrors = 32
```
vyzo: this might even be too many; maybe 16 or 10?

raulk:
For usability, it would be cool to:
- include the value of `maxDialErrors` inside the message for `ErrTooManyErrors`, e.g. "too many errors (max tracked: 32)".
- turn `ErrTooManyErrors` into a struct that counts _how many more errors_ we found beyond our max; when we reach the threshold, we can retain a pointer to it and keep incrementing the counter for every additional error. That way we could at least tell the user how many more errors we saw but failed to retain.

Stebalien:
Alternative here: https://github.com/libp2p/go-libp2p-swarm/pull/120.

Stebalien:
(closing in favor of the alternative, we can reopen if necessary)

PR #121, opened 2019-04-24
dial: return a nice custom dial error

Stebalien:
* Limits the number of dial errors we keep
* Exposes the dial errors
* Avoids string formatting unless we actually need to.

Helps address https://github.com/libp2p/go-libp2p-swarm/issues/119
Alternative to: https://github.com/libp2p/go-libp2p-swarm/pull/120

### DIFF
--- a/dial_error.go
+++ b/dial_error.go
@@ -1,0 +1,69 @@
+package swarm
+
+import (
+	"fmt"
+	"strings"
+
+	peer "github.com/libp2p/go-libp2p-peer"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// maxDialDialErrors is the maximum number of dial errors we record
+const maxDialDialErrors = 16
+
+// DialError is the error type returned when dialing.
+type DialError struct {
+	Peer       peer.ID
+	DialErrors []TransportError
+	Cause      error
+	Skipped    int
+}
+
+func (e *DialError) recordErr(addr ma.Multiaddr, err error) {
+	if len(e.DialErrors) >= maxDialDialErrors {
+		e.Skipped++
+		return
+	}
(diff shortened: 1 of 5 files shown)


dial_error.go
```diff
@@ -0,0 +1,69 @@
...
+		return e.Cause
+	}
+	return nil
+}
+
+var _ error = (*DialError)(nil)
+
+// TransportError is the error returned when dialing a specific address.
+type TransportError struct {
+	Address ma.Multiaddr
```
raulk: Would it make sense to save the `string` representation of the multiaddr to avoid dereferencing and the interface cost? We're likely going to `String()` the maddr for logging anyway.

Stebalien: Probably not. The cost of actually formatting the multiaddr will be higher and, hopefully, we'll only do that once anyways.

Really, we should _avoid_ formatting any of this unless debug logging is enabled but that's going to involve ripping out or fixing our current logging framework.

swarm_dial.go
```diff
@@ -307,10 +313,18 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
 
 	// try to get a connection to any addr
-	connC, err := s.dialAddrs(ctx, p, goodAddrsChan)
-	if err != nil {
-		logdial["error"] = err.Error()
-		return nil, err
+	connC, dialErr := s.dialAddrs(ctx, p, goodAddrsChan)
+	if dialErr != nil {
+		logdial["error"] = dialErr.Cause.Error()
+		if dialErr.Cause == context.Canceled {
```
Stebalien: This fixes a bug introduced by the multierror patch.

vyzo: Nice catch.

PR #129, opened 2019-06-03
feat: cache interface addresses for 1 minute

Kubuxu:
This can be quite an overhead in cases of high connection rates.
The main overhead is thread blocking syscall causing a lot of context
switching.

### DIFF
--- a/swarm_addr.go
+++ b/swarm_addr.go
@@ -1,6 +1,8 @@
 package swarm
 
 import (
+	"time"
+
 	addrutil "github.com/libp2p/go-addr-util"
 	ma "github.com/multiformats/go-multiaddr"
 )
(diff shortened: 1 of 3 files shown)


Kubuxu:
Syscall to fetch all physical interface address called on every request for a swarm address list.

vyzo:
why not cache the addresses completely?

Kubuxu:
We can, the question is if we want to do it.
Both listeners and interface addresses are quite stable.

Kubuxu:
Updated it to cache whole `InterfaceListenAddresses`.
If it comes up on profiles ever again we can swap it out with atomic values or channel generator.

(RWlocks have a nasty property of contending on one cache line, in cpp they sometimes use distributed RWLocks where reader counts are distributed across many cache lines).

swarm_addr.go
```diff
@@ -9,16 +11,58 @@ import (
...
-	return addrutil.ResolveUnspecifiedAddresses(s.ListenAddresses(), nil)
+	s.listeners.RLock() // RLock start
+
+	ifaceListenAddres := s.listeners.ifaceListenAddres
+	isEOL := time.Now().After(s.listeners.cacheEOL)
+	s.listeners.RUnlock() // RLock end
+
+	if ifaceListenAddres != nil && !isEOL {
+		// Cache is valid
+		return ifaceListenAddres, nil
```
Stebalien: We may want to check if our listeners have changed. I'm a bit worried users will call `Listen` and then wonder why that address isn't listed here.

Kubuxu: Done.

Kubuxu:
Fixed.

Kubuxu:
This reduced time for the bootstrap of a dht test network (using https://github.com/Kubuxu/go-libp2p-faketransport) by half.

PR #133, opened 2019-07-09
don't call the "connection handler" for outbound connections

Stebalien:
"oops"

This was never supposed to happen. However, it looks like this has been happening forever. Should we just update the documentation?

If we do that, we should fix the basic host to not call identify from `Connect`.

### DIFF
--- a/swarm_listen.go
+++ b/swarm_listen.go
@@ -88,7 +88,7 @@ func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
 			s.refs.Add(1)
 			go func() {
 				defer s.refs.Done()
-				_, err := s.addConn(c, network.DirInbound)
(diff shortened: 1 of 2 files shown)


swarm_listen.go
```diff
@@ -98,6 +98,14 @@ func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
 					log.Warningf("add conn %s failed: ", err)
 					return
 				}
+
+				// TODO: Get rid of this. We use it for identify
+				// but that should happen much earlier (really,
+				// inside the transport and, if not then, during
+				// the notifications).
+				if h := s.ConnHandler(); h != nil {
```
gap892003, line 106: Is the TODO still valid ?

marten-seemann:
We removed the `ConnHandler` in #286.

PR #138, opened 2019-09-13
Test Swarm Peerstore goroutine leak

aschmahmann:
The test swarm currently leaks goroutines by creating an in memory peerstore and not closing it.

As mentioned in libp2p/go-libp2p#718 we may want to have the swarm close the peerstore when the swarm gets closed, or at the very least the in memory peerstore should take a context just like the datastore peerstore does.

This circumvents the problem for now by just adding another goroutine that kills off the peerstore when the context passed into the test swarm creation is cancelled.

### DIFF
--- a/testing/testing.go
+++ b/testing/testing.go
@@ -73,6 +73,13 @@ func GenSwarm(t *testing.T, ctx context.Context, opts ...Option) *swarm.Swarm {
 	ps.AddPrivKey(p.ID, p.PrivKey)
 	s := swarm.NewSwarm(ctx, p.ID, ps, metrics.NewBandwidthCounter())
 
+	// Close the peerstore when context is expired
+	// TODO: the peerstore should really take a context instead
+	go func() {
+		<-ctx.Done()
+		ps.Close()
+	}()
+
 	tcpTransport := tcp.NewTCPTransport(GenUpgrader(s))
 	tcpTransport.DisableReuseport = cfg.disableReuseport
 


Stebalien:
https://github.com/libp2p/go-libp2p-swarm/pull/139?

We keep abusing contexts. The user can call `Close` _without_ canceling the context and that should still work.

aschmahmann:
I guess, whatever gets the job done. However, from a semantics/style perspective using the context says "whenever this test swarm is done kill the peerstore", where as the other says "swarm close should kill the peerstore".

The context approach is less opinionated on https://github.com/libp2p/go-libp2p/issues/718#issue-490069821 and whether `Close` should kill the peerstore. If you think the `Close` method is better though then no complaints from me.

bigs:
for testing i'm okay with this, though as i think of it... could we add an event to the eventbus and use that?

aschmahmann:
@bigs which direction were you planning on using the eventbus for?

This solution: Use the context (+a goroutine) to close the peerstore
#139: Use Swarm Close to close the peerstore
Eventbus: Which mechanism is registering an event, and when is it being registered?

As I mentioned either this or #139 are fine by me so if you have a preference lmk. Either way I'd like to get one of the fixes merged soon so that the go-libp2p-pubsub tests stop failing from too many goroutines.

Stebalien:
I don't think the event bus is related. IMO, this should be handled with something like uber's fx.

@aschmahmann I prefer not to use contexts to close services because:

1. There's no way to control shutdown order.
2. There's no way to _wait_ for shutdown. This isn't important in this case but is critical for things that need to save stat to datastores.

Stebalien:
I went ahead and merged #139 as it handles the case where we call `Close` on the swarm without canceling the context.

PR #144, opened 2019-10-29
Fix/remove deprecated logging

lanzafame:
Part of the work to unblock: ipfs/go-log#65


Stebalien:
We should probably cut an actual go-log _release_ for this, right?

lanzafame:
@Stebalien https://github.com/ipfs/go-log/issues/68

marten-seemann:
Closing due to age. Looks like https://github.com/ipfs/go-log/pull/65 was merged anyway.

PR #146, opened 2019-11-28
Add Peer Limit to Swarm

nisdas:
This PR adds a peer limit to the Swarm. The purpose of this is to limit new connections/streams being opened with peers once the limit is reached. For my usecase I needed to enforce a peer limit so that my libp2p host would not be connected to more peers than specified. However I could not find a way to configure and enforce this with the current swarm object. If there is a better way to achieve this I would be all ears.

For your reference, this is another branch that adds in a Peer Limit Option, in `go-libp2p` 
https://github.com/libp2p/go-libp2p/compare/master...nisdas:peerLimiter


aarshkshah1992:
@nisdas 

I think [connection manager](https://github.com/libp2p/go-libp2p-connmgr) is what you are looking for. Here's a code sample I used in one of my projects

```
connManager := connmgr.NewConnManager(nMinConnections, nMaxConnections)
opts := []libp2p.Option{
  libp2p.ConnectionManager(connManager),
}
host, err := libp2p.New(ctx, opts...)
```

Let me know if you have any questions.

nisdas:
Unfortunately, connection manager does not satisfy my usecase, its a reactive solution to what I need. It still opens connections with peers and only trims them at every interval. This leads to a lot of peer churn which destabilizes the network, I am trying to avoid connecting to any new peers completely and adding them to my peerstore.

prestonvanloon:
If it’s helpful, we also filed a question on the forum a few days ago. 

https://discuss.libp2p.io/t/what-is-the-proper-way-to-limit-connections/357

swarm.go
```diff
@@ -178,6 +182,10 @@ func (s *Swarm) Process() goprocess.Process {
 }
 
 func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
+	numOfPeers := len(s.conns.m)
+	if numOfPeers >= int(s.peerLimit) && s.Connectedness(tc.RemotePeer()) != network.Connected && s.peerLimit != 0 {
+		return nil, errors.New("number of peers over the peer limit, so rejecting connection")
+	}
```
vyzo: Two things wrong with this:
- you need a default of 0 to mean no peer limit, so that thing work as they currently do (this should be the default)
- you are leaking the connection here, you need to close it if you are rejecting.

nisdas: Hey the if the peer limit is set to zero it ignores this conditional. 
```go 
&& s.peerLimit != 0
```
 I added a var to make this more clear. 
I also now close the connection in the event of an error. 

If I want to set a fixed peer limit , after which no peer will be able to connect with us what would be the best way to go about this ?

vyzo: Put the peer limit not being 0 first please, this was hard to parse.
Also, it's the default, so it will be faster to short-circuit.

In addition, you should log something about this connection closure.

swarm.go
```diff
@@ -181,6 +188,12 @@ func (s *Swarm) Process() goprocess.Process {
 }
 
 func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
+	numOfPeers := len(s.conns.m)
+	nonZeroLimit := s.peerLimit > 0
+	if nonZeroLimit && numOfPeers >= int(s.peerLimit) && s.Connectedness(tc.RemotePeer()) != network.Connected {
+		tc.Close()
```
vyzo: Please log this.
Also, what's the purpose of the Connectedness check?

nisdas: Added in a log, the purpose of the Connectedness check is to continue accepting connections from already connected peers. We only reject connections from peers to which we have no active connections with.

vyzo: but you are not counting peers, you are counting connections.

nisdas: unless I am misunderstanding the code 
```go
len(s.conns.m)
```
wouldn't this represent all the peers with active connections currently ?

The type for `s.conns.m` is  
`map[peer.ID][]*Conn`

vyzo: oh that's true; sorry, my bad.

swarm.go
```diff
@@ -181,6 +188,17 @@ func (s *Swarm) Process() goprocess.Process {
...
 
 func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
+	p := tc.RemotePeer()
+	numOfPeers := len(s.conns.m)
+	nonZeroLimit := s.peerLimit > 0
+
+	// Check if the connection would exceed our specified peer limit.
+	if nonZeroLimit && numOfPeers >= int(s.peerLimit) && s.Connectedness(p) != network.Connected {
+		tc.Close()
+		log.Debugf("rejecting connection with peer [%s]", p)
```
vyzo: nit: "from peer", no need to put the peer ID in brackets.

swarm.go
```diff
@@ -40,6 +40,9 @@ var ErrSwarmClosed = errors.New("swarm closed")
 // transport is misbehaving.
 var ErrAddrFiltered = errors.New("address filtered")
 
+// ErrPeerLimitExceeded is returned when we exceed the specified peer limit.
+var ErrPeerLimitExceeded = errors.New("number of peers over the peer limit, so rejecting connection")
```
vyzo: nit: the "so" is unnecessary, there is no conversation happening.

swarm.go
```diff
@@ -97,12 +103,13 @@ type Swarm struct {
 }
 
 // NewSwarm constructs a Swarm
-func NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter) *Swarm {
+func NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, peerLimit int) *Swarm {
```
vyzo: This is probably best done with functional options, so that we don't break existing code.

nisdas: Any idea on how to make that happen ? I couldn't find a swarm 'config' that I could use instead, so I had to add it as an additional argument.

vyzo: You can make the options take the swarm pointer itself.
So:
```
type Option func(s *Swarm) error
func NewSwarm(stuff, opt Option...)
```

nisdas:
@vyzo addressed your comments and also added in a test case, let me know what you think

swarm_options.go
```diff
@@ -0,0 +1,19 @@
+package swarm
+
+// Option is a Swarm Option that can be given to a Swarm Constructor(`NewSwarm`).
+type Option func(s *Swarm)
```
vyzo: let's also return an error from the option.

swarm_options.go
```diff
@@ -0,0 +1,19 @@
...
+
+func (s *Swarm) ApplyOptions(opts ...Option) {
+	for _, opt := range opts {
+		if opt == nil {
```
vyzo: that seems complete unnecessary.

swarm_options.go
nisdas: the reason I didn't return an error was because if we did, then the
swarm constructor also has to return an error
```go
 NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, opts ...Option) (*Swarm,error) {
```
This would break existing code, if you still want me to change, I will 
add it in

vyzo: Ah yes, good point; let's not.

vyzo: or we can panic on error, but that's something we can do from the option itself I guess.

swarm_test.go
```diff
@@ -61,7 +61,7 @@ func makeDialOnlySwarm(ctx context.Context, t *testing.T) *Swarm {
 	return swarm
 }
 
-func makeSwarms(ctx context.Context, t *testing.T, num int, opts ...Option) []*Swarm {
+func makeSwarms(ctx context.Context, t *testing.T, num int, opts ...TestOption) []*Swarm {
```
vyzo: what's this `TestOption`?

vyzo: ah found it, nm.

vyzo:
travis test failure because of a data race.

nisdas:
@vyzo fixed the data race, sounds good I will be waiting for @Stebalien review

nisdas:
@raulk 
Hey thanks for the detailed reply, looking through the code some more you are right that this is still a reactive solution. It seems like `addCon` is called after the connections are initiated so it would essentially be a more extreme version of the connection manager. 

For outgoing connections, it is simple enough to put this check right before we dial a peer https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm_dial.go#L293 . This would prevent any new outgoing connection from being created entirely. However for inbound connections this is a bit more complicated as there is no way to prevent an external peer from dialing us, from what I understand of the code the best place we can probably put a limit check is here https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm_listen.go#L79. 

My main purpose for this PR was to prevent any connection notification handlers from being called if we exceed the peer limit. https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm.go#L235 . Since this lead to a lot of the peer churn issues we were seeing when using the connection manager previously. 

I will look into implementing a much more strict connection manager for this, since it looks like adding this PR in might bring about some security risks to the protocol.

nisdas:
Closing this PR as we have chosen to take an alternative approach of using a strict connection manager over here https://github.com/prysmaticlabs/prysm/pull/4110 .

PR #157, opened 2020-01-22
 Introspector in swarm

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/775

Changes for swarm to be able to register as a data provider & then provide the data/metrics when it is asked to do so.


aarshkshah1992:
### **TODO**
- [ ] Release go-libp2p-core & configure the version here
- [ ] Build the provider for connections & stream
- [ ] Tests for the providers

swarm.go
```diff
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	coreit "github.com/libp2p/go-libp2p-core/introspection"
```
raulk: Shorten `introspection` to `introspect`, and drop the import alias.

swarm.go
```diff
@@ -115,6 +116,13 @@ func NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc
 	s.proc = goprocessctx.WithContextAndTeardown(ctx, s.teardown)
 	s.ctx = goprocessctx.OnClosingContext(s.proc)
 
+	// TODO register provider for introspection
+	if introspector != nil {
```
raulk: Would love to see the actual provider functions, after my comments on https://github.com/libp2p/go-libp2p-core/pull/107 have been integrated.

aarshkshah1992:
@raulk  Here are the fields that still need some discussion before we implement sourcing for them:

### Stream

1.  **UserProvidedTags**
    What does this mean ?
2.

conn_introspect.go
```diff
@@ -0,0 +1,95 @@
...
+func (ci *connIntrospector) id() string {
+	return ci.c.id
+}
+
+func (ci *connIntrospector) remotePeerId() string {
+	return ci.c.RemotePeer().String()
+}
+
+func (ci *connIntrospector) endPoints() *introspectpb.EndpointPair {
+	return &introspectpb.EndpointPair{SrcMultiaddr: ci.c.LocalMultiaddr().String(), DstMultiaddr: ci.c.RemoteMultiaddr().String()}
```
raulk: nit: multi-line init is more readable in this case.

conn_introspect.go
```diff
@@ -0,0 +1,95 @@
...
+		return introspectpb.Role_RESPONDER
+	} else {
+		return introspectpb.Role_INITIATOR
+	}
+}
+
+// TODO Number of packets & instantaneous bandwidth ?
+// Should we have a separate "flow-metre" for a connection to prepare for a message oriented world ?
+func (ci *connIntrospector) traffic() *introspectpb.Traffic {
+	if ci.s.bwc != nil {
```
raulk: In these cases, prefer short-circuiting to reduce one level of nesting/indentation.

```go
if ci.s.bwc == nil {
  return nil
}
<rest>
```

introspect.go
```diff
@@ -0,0 +1,173 @@
...
+
+import (
+	"github.com/libp2p/go-libp2p-core/introspect"
+	introspectpb "github.com/libp2p/go-libp2p-core/introspect/pb"
+	"github.com/pkg/errors"
+)
+
+// IntrospectTraffic introspects & returns the overall traffic for this peer
+func (s *Swarm) IntrospectTraffic() (*introspectpb.Traffic, error) {
+	if s.bwc != nil {
```
raulk: Reverse the condition, and outdent the logic outside the if.

introspect.go
```diff
@@ -0,0 +1,173 @@
...
+		t := &introspectpb.Traffic{}
+		metrics := s.bwc.GetBandwidthTotals()
+
+		t.TrafficIn = &introspectpb.DataGauge{CumBytes: uint64(metrics.TotalIn), InstBw: uint64(metrics.RateIn)}
```
raulk: nit: multi-line init is more readable.

introspect.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return nil, nil
+}
+
+// IntrospectConnections introspects & returns the swarm connections
+func (swarm *Swarm) IntrospectConnections(query introspect.ConnectionQueryParams) ([]*introspectpb.Connection, error) {
+	swarm.conns.RLock()
+	defer swarm.conns.RUnlock()
+
+	containsId := func(ids []introspect.ConnectionID, id string) bool {
```
raulk: More efficient to create a `map[introspect.ConnectionID]struct{}`, and make the predicate test membership. This is `O(m*n)` worst case scenario; map-based solution would be `O(m)` at the expense of a little extra memory.

introspect_test.go
```diff
@@ -0,0 +1,129 @@
+package swarm_test
+
+import (
+	"context"
```
raulk: Regroup imports.

swarm.go
```diff
@@ -209,11 +222,13 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	}
 
 	// Wrap and register the connection.
+	id := uuid.NewV4()
```
raulk: An UUID is pretty inefficient. Requires string comparison. Why not an int32 atomic counter at the swarm level, which you increment with every new connection?

aarshkshah1992:
@raulk I'm happy to refactor this PR along with implementing the stream/connection scoped metering if you like.

PR #164, opened 2020-02-26
fix: fire a listen close event when closing the listener

Stebalien:
fixes https://github.com/libp2p/go-libp2p-swarm/issues/163

### DIFF
--- a/swarm_notif_test.go
+++ b/swarm_notif_test.go
@@ -16,18 +16,32 @@ import (
 func TestNotifications(t *testing.T) {
 	const swarmSize = 5
 
+	notifiees := make([]*netNotifiee, swarmSize)
+
 	ctx := context.Background()
 	swarms := makeSwarms(ctx, t, swarmSize)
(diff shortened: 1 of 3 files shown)


swarm_notif_test.go
```diff
@@ -16,18 +16,32 @@ import (
...
+			select {
+			case <-notifiees[i].listenClose:
+				t.Error("should not have been closed")
+			default:
+			}
+			err := s.Close()
+			if err != nil {
+				t.Error(err)
+			}
+			select {
```
aarshkshah1992: Should _probably_ fail the test on a timed context rather than the default case as the `ListenClose ` notif is sent inside a go-routine that `Swarm.Close` does not synchronise with.

aarshkshah1992: I take my word back. `Swarm.Close` waits of all refs to complete, one of which is completed by the go-routine that sends the `ListenClose ` notification.

Apporved.

PR #166, opened 2020-02-28
Emit swarm events on the bus 

aarshkshah1992:
@Stebalien 

For https://github.com/libp2p/go-libp2p/issues/801


Stebalien:
What if we just did this in the host? It might be simpler to just listen for low-level events there and turn them into higher-level event-bus events.

aarshkshah1992:
@Stebalien 

That's a very good idea. Considering we always want to keep this backward compatible, we're never going to remove the Swarm Notifiee anyways & so it's not a bad idea to hook onto it in the host to make the wiring of this easier/remove the messy construction logic in both the host and swarm.

aarshkshah1992:
Closing this PR as we have decided to make the Host emit the network events on the bus. Related PR at https://github.com/libp2p/go-libp2p/pull/813.

Stebalien:
> That's a very good idea. Considering we always want to keep this backward compatible, we're never going to remove the Swarm Notifiee anyways & so it's not a bad idea to hook onto it in the host to make the wiring of this easier/remove the messy construction logic in both the host and swarm.

Note: we'll likely merge the swarm/host in some major refactor later. But not for a long time.

PR #185, opened 2020-03-30
feat: handle no addresses

Stebalien:
If we aren't listening on any addresses, we should return no addresses, not an error.


willscott:
hopefully, fix the broken test / file an issue:
```
--- FAIL: TestSimultDials (0.02s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
```

the change seems functionally safe, although i was suspicious enough of indexing into a nil slice that I made sure i could compile/run that line without a panic.

Stebalien:
That's definitely a bug.

PR #201, opened 2020-04-07
implement connection gating support: intercept peer, address dials, upgraded conns

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/872.

Core PR at https://github.com/libp2p/go-libp2p-core/pull/139.
Upgrader PR at https://github.com/libp2p/go-libp2p-transport-upgrader/pull/55.

This PR uses the connection lifecycle state based Connection Gating interface to gate connections.

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -4,12 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/libp2p/go-libp2p-core/connmgr"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/metrics"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
--- a/swarm.go
+++ b/swarm.go
@@ -182,7 +182,21 @@ func (s *Swarm) Process() goprocess.Process {
 }
 
 func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
-	p := tc.RemotePeer()
+	var (
+		p    = tc.RemotePeer()
+		addr = tc.RemoteMultiaddr()
+	)
+
+	if s.gater != nil {
+		if allow := s.gater.InterceptAddrDial(p, addr); !allow {
+			err := tc.Close()
+			if err != nil {
+				log.Warnf("failed to close connection with peer %s and addr %s; err: %s", p.Pretty(), addr, err)
+			}
+			return nil, ErrAddrFiltered
+		}
+	}
+
 	stat := network.Stat{Direction: dir}
 	c := &Conn{
 		conn:  tc,
@@ -195,11 +209,11 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	if s.gater != nil {
 		if allow, _ := s.gater.InterceptUpgraded(c); !allow {
 			// TODO Send disconnect with reason here
-			if err := tc.Close(); err != nil {
-				log.Errorf("failed to close connection with peer %s and addr %s; err: %s",
-					p.Pretty(), tc.RemoteMultiaddr(), err)
+			err := tc.Close()
+			if err != nil {
+				log.Warnf("failed to close connection with peer %s and addr %s; err: %s", p.Pretty(), addr, err)
 			}
-			return nil, ErrConnectionAttemptGated
+			return nil, ErrGaterDisallowedConnection
 		}
 	}
 
--- a/swarm.go
+++ b/swarm.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/libp2p/go-libp2p-core/connmgr"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -19,9 +20,7 @@ import (
 	"github.com/jbenet/goprocess"
 	goprocessctx "github.com/jbenet/goprocess/context"
 
-	filter "github.com/libp2p/go-maddr-filter"
 	ma "github.com/multiformats/go-multiaddr"
-	mafilter "github.com/whyrusleeping/multiaddr-filter"
 )
 
 // DialTimeoutLocal is the maximum duration a Dial to local network address
@@ -88,8 +87,7 @@ type Swarm struct {
 	backf   DialBackoff
 	limiter *dialLimiter
 
-	// filters for addresses that shouldnt be dialed (or accepted)
-	Filters *filter.Filters
+	ConnGater connmgr.ConnectionGater
 
 	proc goprocess.Process
 	ctx  context.Context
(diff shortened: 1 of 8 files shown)


swarm.go
```diff
@@ -168,32 +165,27 @@ func (s *Swarm) teardown() error {
...
 }
 
-func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
-	// The underlying transport (or the dialer) *should* filter it's own
-	// connections but we should double check anyways.
-	raddr := tc.RemoteMultiaddr()
-	if s.Filters.AddrBlocked(raddr) {
-		tc.Close()
-		return nil, ErrAddrFiltered
+func (s *Swarm) denyConn(tc transport.CapableConn, dir network.Direction) (deny bool) {
```
willscott, line 173: given that there's only one caller of this function, and the way that the ConnGater is getting invoked is slighty different based on if we know the specific address or just the peer, i wonder if this function is useful, or if we should just directly check the ConnGater in swarm_listen, like we do when dialing.

testing/testing.go
```diff
@@ -99,3 +113,57 @@ func DivulgeAddresses(a, b network.Network) {
 	addrs := a.Peerstore().Addrs(id)
 	b.Peerstore().AddAddrs(id, addrs, peerstore.PermanentAddrTTL)
 }
+
+// MockConnectionGater is a mock connection gater to be used by the tests.
```
willscott, line 117: This seems potentially useful as a test helper that could be re-used.
I'm not sure if I'd look for it here. Probably only after checking go-libp2p/p2p/net/mock and near go-libp2p-core/connmgr

aarshkshah1992: Hey @willscott 

I agree. Since this interface and PR is currently heavy WIP, I'd like to keep it here till it lands and then pull it out to /net/mock. Have added a TODO item for this to https://github.com/libp2p/go-libp2p/issues/872 to track it.

swarm.go
```diff
@@ -168,33 +165,46 @@ func (s *Swarm) teardown() error {
...
+	// last fallback check for gating connections
+	var rejectConnection bool
+	if s.ConnGater != nil {
+		switch dir {
+		case network.DirInbound:
+			if !s.ConnGater.InterceptAccept(tc) {
+				rejectConnection = true
+			}
+		case network.DirOutbound:
+			if !s.ConnGater.InterceptPeerAddrDial(p, tc.RemoteMultiaddr()) || !s.ConnGater.InterceptPeerDial(p) {
```
raulk, line 184: Why do we intercept this after the connection has been established? Can't we intercept it before we make the connection attempt (since we already know the target peer ID and the multiaddr we're about to dial)?

swarm.go
```diff
@@ -168,33 +165,46 @@ func (s *Swarm) teardown() error {
...
+				rejectConnection = true
+			}
+		}
+
+		if !s.ConnGater.InterceptSecured(dir, p, tc) {
+			rejectConnection = true
+		}
```
raulk, line 191: For transports that use the upgrader, this will end up invoking the gater twice. Wouldn't it be simpler to modify the QUIC transport? We can't have two layers (transport and network) being responsible for doing the same thing; it's a hack.

swarm_dial.go
```diff
@@ -409,7 +418,9 @@ func (s *Swarm) filterKnownUndialables(addrs []ma.Multiaddr) []ma.Multiaddr {
 		s.canDial,
 		// TODO: Consider allowing link-local addresses
 		addrutil.AddrOverNonLocalIP,
-		addrutil.FilterNeg(s.Filters.AddrBlocked),
+		func(addr ma.Multiaddr) bool {
+			return s.ConnGater == nil || s.ConnGater.InterceptPeerAddrDial(p, addr)
```
raulk, line 422: Why do this here and on `addConn`?

swarm_dial.go
```diff
@@ -312,6 +315,12 @@ func (s *Swarm) canDial(addr ma.Multiaddr) bool {
 
 // dial is the actual swarm's dial logic, gated by Dial.
 func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
+	// should we allow the connection to this peer at all ?
+	if s.ConnGater != nil && !s.ConnGater.InterceptPeerDial(p) {
+		log.Debugf("gater disallowed outbound connection to peer %s", p)
```
raulk, line 320: Don't ask why...

```suggestion
		log.Debugf("gater disallowed outbound connection to peer %s", p.Pretty())
```

swarm_test.go
```diff
@@ -280,60 +282,102 @@ func TestConnHandler(t *testing.T) {
...
-	})
+	for n, tc := range tcs {
+		p1Gater := DefaultMockConnectionGater()
+		p2Gater := DefaultMockConnectionGater()
+		if tc.p1Gater != nil {
+			p1Gater = tc.p1Gater(p1Gater)
+		}
+		if tc.p2Gater != nil {
+			p2Gater = tc.p2Gater(p2Gater)
+		}
```
raulk, line 364: Can we use `t.Run()` here to make these actual subtests instead of one huge test with many cases?

swarm_dial.go
aarshkshah1992: @raulk Have moved this check to `Swarm.DialPeer` and have changed it to p.Pretty.

swarm_dial.go
aarshkshah1992: This has been removed from `addConn`.

swarm.go
aarshkshah1992: This has been removed.

swarm.go
aarshkshah1992: This has been removed. Let's add it to the QUIC transport. The meta-issue has a linked PR to do the reflection magic to inject Gaters into Transports and also has a TODO to modify the QUIC transport code to do this.

swarm_test.go
aarshkshah1992: Good idea. Done.

aarshkshah1992:
@raulk Have changed the imports  for libp2p-core & libp2p-upgrader & have addressed your changes. Let me know what you think.

swarm.go
raulk: Probably the same applies to `InterceptUpgraded`, but let's keep it for now.

swarm.go
```diff
@@ -168,33 +176,32 @@ func (s *Swarm) teardown() error {
...
-	raddr := tc.RemoteMultiaddr()
-	if s.Filters.AddrBlocked(raddr) {
-		tc.Close()
-		return nil, ErrAddrFiltered
+	p := tc.RemotePeer()
+	stat := network.Stat{Direction: dir}
+	c := &Conn{
+		conn:  tc,
+		swarm: s,
+		stat:  stat,
```
raulk, line 190: I really hate this, but I'm going to have to re-add the call to `InterceptAddrDial` to avoid causing a regression.

swarm.go
```diff
@@ -182,7 +182,21 @@ func (s *Swarm) Process() goprocess.Process {
...
 
 func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
-	p := tc.RemotePeer()
+	var (
+		p    = tc.RemotePeer()
+		addr = tc.RemoteMultiaddr()
+	)
+
+	if s.gater != nil {
+		if allow := s.gater.InterceptAddrDial(p, addr); !allow {
```
aarshkshah1992, line 191: @raulk Why do we need this here ?

raulk: https://github.com/libp2p/go-libp2p-swarm/pull/201#discussion_r425792282

PR #202, opened 2020-04-12
Initialize extra map in connections and streams Stat

pymq:
Currently it seems like `Stat() Stat` API is broken - `Extra` map is always nil and devs have no possibility to use it. I am not sure is it a good idea to spend a few extra bytes on every stream, but this is talk for another PR I suppose, where we can discuss more flexible API.


Stebalien:
Thanks but this map was never intended to be a place for users to stash information, there's no safe/non-racy way to do that. It was intended to _eventually_ expose additional transport-specific information from  the underlying stream multiplexer (but that was never implemented).

PR #212, opened 2020-05-05
Rank Dial addresses

aarshkshah1992:
* A dial address is FD consuming when:
  * For non-relay addresses, the transport is TCP or WebSocket.
  * For relay addresses, the transport of the _Relay Server_ is TCP or Websocket.
* The limiter ONLY consumes FD's for non-relay addresses to avoid double counting.
* For dialling:
  * We partition the addresses into FD consuming and Non-FD consuming addresses. 
  * We then sort each of two sets in descending order of priority such that private addresses have the highest priority followed by non-relay public addresses followed by relay addresses.
* We first attempt to dial all the sorted Non-FD consuming addresses and ONLY after/if we exhaust all of them without getting a connection, we attempt to dial the sorted FD consuming addresses.


swarm_dial.go
```diff
@@ -529,3 +593,48 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
 	// success! we got one!
 	return connC, nil
 }
+
+// TODO We should have a `IsFdConsuming() bool` method on the `Transport` interface in go-libp2p-core/transport.
```
Stebalien, line 597: Probably, but later.

swarm_dial.go
```diff
@@ -529,3 +593,48 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
...
+// This function checks if any of the transport protocols in the address requires a file descriptor.
+func (s *Swarm) IsFdConsumingAddr(addr ma.Multiaddr) bool {
```
Stebalien, line 599: I wouldn't export this.

swarm_dial.go
```diff
@@ -343,23 +343,87 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+	fdAddrs = rankAddrsFnc(fdAddrs)
+
 	/////////
+	// first try to get a connection to a non-fd consuming addr and if that fails,
+	// try to get a connection to a fd consuming addr.
+	var connC transport.CapableConn
+	var errNonFd *DialError
+	var dialErr *DialError
+
+	connC, errNonFd = s.dialAddrs(ctx, p, addrsToChanFnc(nonFdAddrs))
```
Stebalien, line 417: We _can_ just get rid of the channel on this interface to simplify it.

swarm_dial.go
```diff
@@ -343,23 +343,87 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+	if errNonFd != nil {
+		var errFd *DialError
+		connC, errFd = s.dialAddrs(ctx, p, addrsToChanFnc(fdAddrs))
+		if errFd != nil {
+			dialErr = &DialError{Peer: p, DialErrors: append(errNonFd.DialErrors, errFd.DialErrors...),
```
Stebalien, line 422: We should drop additional errors if we're over the `maxDialDialErrors` limit (not sure what I did with that variable name). We might want to just add a "combine" method to the `DialError` type.

swarm_dial.go
```diff
@@ -343,23 +343,87 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+		close(goodAddrsChan)
+		return goodAddrsChan
+	}
+
+	// ranks addresses in descending order of preference for dialing
+	// local addrs > non-relay public addrs > relay-addrs
+	rankAddrsFnc := func(addrs []ma.Multiaddr) []ma.Multiaddr {
+		var localAddrs []ma.Multiaddr
+		var relayAddrs []ma.Multiaddr
+		var others []ma.Multiaddr
```
Stebalien, line 391: I'd pre-allocate and over-allocate (i.e., with length `len(addrs)`) all of these.

swarm_dial.go
```diff
@@ -529,3 +593,48 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
...
+		relayaddr, _ := ma.SplitFunc(addr, func(c ma.Component) bool {
+			return c.Protocol().Code == ma.P_CIRCUIT
+		})
+		if relayaddr == nil {
+			log.Debugf("proxy address %s did not have a Circuit Component,"+
+				"so marking it as FD Consuming for safety", addr.String())
+			return true
+		}
+		return noProxyAddrFnc(relayaddr)
+	}
```
Stebalien, line 635: This isn't necessary. The relay transport doesn't consume a file descriptor.

When the relay transport dials the relay, it'll dial that relay with the swarm. When it does, the swarm will take any fd tokens necessary based on the transport.

swarm_dial.go
```diff
@@ -529,3 +593,48 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
...
+		}
+		protos := tpt.Protocols()
+		// is a fd consuming transport present in the list of protocols that the transport supports ?
+		for i := range protos {
+			if _, ok := fdConsumingTptProtos[protos[i]]; ok {
+				return true
+			}
+		}
+		return false
+	}
```
Stebalien, line 615: Hm. I take back what I said before. Let's just simplify this:

1. If the transport is a "proxy" transport, don't consume a file descriptor token.
2. Otherwise, just call `addrutil.IsFDCostlyTransport`.

Stebalien (changes requested):

swarm_dial.go
Stebalien: Note: as the rules here grow more complicated, it may be simpler to use `sort.Slice()`.

swarm_dial.go
aarshkshah1992: Agreed.

swarm_dial.go
aarshkshah1992: @Stebalien Please can you explain why pre-allocating and especially over-allocating would be better than allocating on the go ? What trade-off does that get us ?

Generally speaking, if we have a maximum bound on the size of a slice, why should we always pre-allocate even if that risks over-allocating rather than just allocating on the go ?

swarm_dial.go
aarshkshah1992: @Stebalien How do I unit-test it without exporting it ? 

1. All the testing helpers are in a `/testing` package which depends on go-libp2p-swarm.
2. All our tests are in `swarm_test` package.

Do you want me to write a unit-test in the swarm package even if that means copying some of the stuff in the `testing` package ?

swarm_dial.go
aarshkshah1992: Agreed. This is done.

swarm_dial.go
aarshkshah1992: @Stebalien  Using `sort.Slice now` and got rid of the allocs.

swarm_dial.go
aarshkshah1992: This is done.

swarm_dial.go
aarshkshah1992: @Stebalien 

I completely understand that. That's why we have the `dialLimiter.shouldConsumeFd` function which does NOT consume a FD for relays.

However, for partitioning the addresses into FD/Non-FD addresses so we can wait on all dials for the latter before dialing the former, I'd like to **classify** relays where the relay server has a FD consuming addresses as FD consuming so we dial them in the "second batch". This function serves that purpose.

swarm_dial.go
aarshkshah1992: @Stebalien Please see above comment.

For dialing in the limiter, we do NOT consume fd's for proxy addrs.
For classifying an address as FD or Non-FD consuming so we can dial all the latter ones before attempting the former, we look at the address of the relay server.

aarshkshah1992:
@Stebalien Have addressed your review with some open comments. Please take a look.

dial_error.go
```diff
@@ -21,6 +21,29 @@ type DialError struct {
...
+	for i := range second.DialErrors {
+		if len(cbd.DialErrors) >= maxDialDialErrors {
+			cbd.Skipped++
+		} else {
+			cbd.DialErrors = append(cbd.DialErrors, second.DialErrors[i])
+		}
+	}
+
+	return cbd
+}
```
Stebalien, line 45: We should be able to do this with two allocations (slice + error), two appends, some some slicing, and no for loops.

swarm_dial.go
Stebalien: I thought I commented on this....

This is fine as-is (I didn't realize why we were exporting it).

swarm_dial.go
Stebalien: Ah, I see. You're right, we do need to do that.

Stebalien: In that case, I'd still simplify this (given that we only really support circuit addresses anyways):

1. Modify `addrutil.IsFDCostlyTransport` to ignore circuit addresses.
2. When sorting between relay connections when dialing, split on the `P_CIRCUIT` protocol and call `addrutil.IsFDCostlyTransport` on the relay part.

That way:

* We don't need to look at the swarm's transports. You're right in that we should probably expose this information on the transport itself, but we don't do that yet anyways.
* Can just go "full special case" for circuit addresses.

Does it sound like that will simplify this?

dial_error.go
aarshkshah1992: @Stebalien Turns out we need only one alloc. This is done.

aarshkshah1992:
@Stebalien We shouldn't leak how our dial uses the `addrutil.IsFDCostlyTransport` utility into the _shared_ utility itself.

I've documented the code better for now and have filed an issue to simplify/fix this later.

Have made the other changes. Please take a look when you can.

swarm_dial.go
```diff
@@ -343,23 +344,77 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+	sortAddrs(fdAddrs)
+
 	/////////
+	// first try to get a connection to a non-fd consuming addr and if that fails,
+	// try to get a connection to a fd consuming addr.
+	var connC transport.CapableConn
+	var errNonFd *DialError
+	var dialErr *DialError
+
+	connC, errNonFd = s.dialAddrs(ctx, p, nonFdAddrs)
```
raulk, line 409: I don't think this is a good idea. QUIC is a UDP-based protocol. UDP is unreliable, it has no connection establishment flow like TCP, the feedback from lack of connectivity/port closed is not immediate like with TCP. If we dialled an IP that was routable but a bad/closed port, I suspect we'd have to wait for the full connection timeout to notice, whereas with TCP we'd notice during the TCP handshake (much sooner). Same applies if firewalls/routers are dropping UDP packets.

What does this buy us? If the ultimate goal is to select a succeeding a QUIC connection over a succeeding TCP connection, sequentialising the dials is going to cause more pain than gain IMO.

raulk (changes requested):

swarm_dial.go
Stebalien: The TCP handshake is a layer below TCP reliability. It's all IP at that point:

* The feedback for a UDP packet sent to a closed port is the same as the feedback for TCP: an ICMP unreachable packet. Unfortunately, many home firewalls will just _drop_ packets so you'll get no feedback either way.
* If a SYN gets dropped, it will be resent eventually (2s?). I assume QUIC does the same.

The real question is whether or not the ICMP unreachable packet this is exposed to the reader. @marten-seemann, any ideas on this?

> What does this buy us?

It:

* Avoids tying up slots in the file descriptor dial limiter needed by peers that only support TCP.
* Avoids spraying SYN packets (NATs don't like that).

However, it's not critical. I'm fine landing a version of this patch that simply prioritizes UDP connections.

marten-seemann: > The TCP handshake is a layer below TCP reliability. It's all IP at that point

That is correct.

> If a SYN gets dropped, it will be resent eventually (2s?). I assume QUIC does the same.

This is the order of magnitute that the TCP RFC recommends, but most implementations use a much shorter value. Similarly, the QUIC draft was changed after discussing with the TCP folks to use the same value, but not many implementations are willing to take that performance hit. quic-go retransmits two copies of the Initial packet after 200ms (with exponential backoff after that).

> The real question is whether or not the ICMP unreachable packet this is exposed to the reader. @marten-seemann, any ideas on this?

For UDP, the kernel will only deliver ICMP packets if you're using a connected socket.

You can't really rely on ICMP packets anyway, since they're frequently dropped (or mis-routed). Furthermore, as ICMP packets are not authenticated, it would be inadvisable to take any action, except for the very early stages of the handshake.

aarshkshah1992:
@steb @raulk We no longer wait for UDP addresses to finish before dialling TCP addresses. Please take a look at the PR when you can.

swarm_dial.go
```diff
@@ -343,23 +344,67 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
-	close(goodAddrsChan)
 	if !nonBackoff {
 		return nil, ErrDialBackoff
 	}
-	/////////
 
-	// try to get a connection to any addr
-	connC, dialErr := s.dialAddrs(ctx, p, goodAddrsChan)
+	// sorts addresses in descending order of preference for dialing
+	// Private UDP > Private TCP > Public UDP > Public TCP > UDP Relay server > TCP Relay server
```
Stebalien, line 361: I'd actually say public UDP goes before private TCP as TCP dials can get blocked on the fd limiter.

swarm_dial.go
Stebalien: @aarshkshah1992 thoughts on this?

swarm_dial.go
aarshkshah1992: That makes sense. Have made the change.

swarm_dial.go
aarshkshah1992: @Stebalien I'd replied earlier at https://github.com/libp2p/go-libp2p-swarm/pull/212#issuecomment-626501013.

Reproducing it here:

"We shouldn't leak how our dial uses the addrutil.IsFDCostlyTransport utility into the 
shared utility itself. I've documented the code better for now and have filed an issue(#214 ) to simplify/fix this later.
".

aarshkshah1992:
@Stebalien Have addressed your review and we now rank Public UDP addresses above private TCP addresses. Please take a look when happy.

aarshkshah1992:
ping @Stebalien for one last look.

swarm_dial.go
```diff
@@ -56,6 +59,11 @@ var (
 	ErrGaterDisallowedConnection = errors.New("gater disallows connection to peer")
 )
 
+var fdConsumingTptProtos = map[int]struct{}{
+	ma.P_WS:  struct{}{},
+	ma.P_TCP: struct{}{},
+}
```
Stebalien, line 65: I'd strongly prefer to just check if the address contains `tcp` and/or `unix` (i.e., a fd consuming transports) before any proxy addresses. Otherwise, we're going to miss transports (e.g., WSS).

swarm_dial.go
Stebalien: I still think this code needs to be simpler:

* We're checking to see if the address is a "proxy" address, but we only _actually_ support circuit addresses. We should just split on the circuit address and be done with it.
* We're hard coding rules for specific transport protocols when we should be able to just say "if protocol X appears in the target multiaddr (before any proxy protocols), the address must consume a file descriptor". That is, we know that the tcp and unix protocols both _must_ consume a file descriptor per connection while udp, memory, etc. protocols don't.

Basically, this code is trying to be general purpose and abstract over transports, but we can't actually do that so it's hard coding a bunch of stuff. Sitting halfway in between is just confusing.

swarm_dial.go
```diff
@@ -352,23 +353,66 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+			// private TCP
+			if isPrivate {
+				return 3
+			}
+			// public tcp
+			if !relay {
+				return 4
+			}
+			return 6
+		}
```
Stebalien, line 401: nit: Numeric scores are hard to read and easy to get wrong. Honestly, now that I see this, I'd go back to sorting these addresses into separate slices then concatenating them like you were doing before (sorry for going back and forth on this).

Note: This _is_ fine as-is, just a bit confusing.

aarshkshah1992: I agree. I think that way is easier to read. This is done.

swarm_dial.go
aarshkshah1992: I think I grok what you mean here and in https://github.com/libp2p/go-libp2p-swarm/pull/212#discussion_r427025170. I've made the changes.

aarshkshah1992:
@Stebalien I've addressed all your concerns and it looks simpler/cleaner now. Please take a look.

swarm_dial.go
```diff
@@ -352,23 +347,60 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+		var localUdpAddrs []ma.Multiaddr // private udp
+		var relayUdpAddrs []ma.Multiaddr // relay udp
+		var othersUdp []ma.Multiaddr     // public udp
+
+		var localFdAddrs []ma.Multiaddr // private fd consuming
+		var relayFdAddrs []ma.Multiaddr //  relay fd consuming
+		var othersFd []ma.Multiaddr     // public fd consuming
+
+		for _, a := range addrs {
+			if manet.IsPrivateAddr(a) {
```
Stebalien, line 375: I think this will include circuit addrs. We should probably check for that _first_.

Stebalien (approved):
One small change. Otherwise, LGTM!

swarm_dial.go
aarshkshah1992: Good catch. This is done.

PR #225, opened 2020-06-24
slim down dependencies

Stebalien:
Remove secio, quic, and circuit transports from testing dependencies.

* We temporarily added quic/circuit dependencies because a test needed them. But this is no longer the case.
* We introduced the secio dependency before we had a fully featured "insecure" transport for testing.


Stebalien:
`IsFdConsumingAddr` is a pure function. This test added the QUIC and circuit transports because, when it was first written, `IsFdConsumingAddr` looked at the registered transports themselves. But that was changed before the PR adding `IsFdConsumingAddr` was merged.

Stebalien:
TL;DR: this doesn't affect code coverage.

PR #227, opened 2020-08-28
feat: close transports that implement io.Closer

### LINKED ISSUE #999: Proposal: Closable Transport
# Use Case
While writing a tor transport I noticed that some transports runs some node beyond the listener (mainly overlay networks).
For example my tor transport runs a tor node and I need to close it at one point. I can't rely on the listeners to close it because it's not because a node stop listening on an address it will not then dial someone.
# Additions
I propose to add this in `go-libp2p-core/transport` :
```go
type ClosableTransport interface {
  Transport

  // Close is called when the libp2p node doesn't need the transport anymore (the is node closing).
  // After that the transport can safely assume no more Listen() or Dial() are gonna be done.
  // This is usefull for transport running overlay network.
  Close() error
}
```
Then in `host.Close()` the host would use reflection to terminate the transport when closing (after closing all listeners) :
```go
// Not actual working code, just an idea
var wg sync.WaitGroup
for _, v := range h.transports {
  ct, ok := v.(tpt.ClosableTransport)
  if ok {
    wg.Add(1)
    go func (t tpt.ClosableTransport){ // Needed because range is not thread safe
      defer wg.Done()
      t.Close()
    }(ct)
  }
}
wg.Wait()
```
(Also I can't rely on the garbage collector to do this work because I have some CGO code (or a system fork) and the garbage collector doesn't deal with that.)
## In the Future
If in the future a method to unhook transport from a libp2p is added the transport will be closable that way.

Stebalien:
This way, transports with shared resources (e.g., reused sockets) can clean them up. Otherwise, we have no way to tell, e.g., the QUIC transport to close it's socket "pool".

fixes https://github.com/libp2p/go-libp2p/issues/999

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -175,6 +176,27 @@ func (s *Swarm) teardown() error {
 
 	// Wait for everything to finish.
 	s.refs.Wait()
+
+	// Now close out any transports (if necessary). Do this after closing
+	// all connections/listeners.
+	s.transports.Lock()
+	transports := s.transports.m
+	s.transports.m = nil
+	s.transports.Unlock()
+
+	var wg sync.WaitGroup
+	for _, t := range transports {
+		if closer, ok := t.(io.Closer); ok {
+			wg.Add(1)
+			go func(c io.Closer) {
+				defer wg.Done()
+				if err := closer.Close(); err != nil {
+					log.Errorf("error when closing down transport %T: %s", c, err)
+				}
+			}(closer)
+		}
+	}
+	wg.Wait()
 
 	return nil
 }
--- a/swarm_listen.go
+++ b/swarm_listen.go
@@ -40,7 +40,17 @@ func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
 func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
 	tpt := s.TransportForListening(a)
 	if tpt == nil {
-		return ErrNoTransport
+		// TransportForListening will return nil if either:
+		// 1. No transport has been registered.
+		// 2. We're closed (so we've nulled out the transport map.
+		//
+		// Distinguish between these two cases to avoid confusing users.
+		select {
+		case <-s.proc.Closing():
+			return ErrSwarmClosed
+		default:
+			return ErrNoTransport
+		}
 	}
 
 	list, err := tpt.Listen(a)
--- a/swarm_transport.go
+++ b/swarm_transport.go
@@ -20,7 +20,10 @@ func (s *Swarm) TransportForDialing(a ma.Multiaddr) transport.Transport {
 	s.transports.RLock()
 	defer s.transports.RUnlock()
 	if len(s.transports.m) == 0 {
-		log.Error("you have no transports configured")
+		// make sure we're not just shutting down.
+		if s.transports.m != nil {
+			log.Error("you have no transports configured")
+		}
 		return nil
 	}
 
@@ -48,7 +51,10 @@ func (s *Swarm) TransportForListening(a ma.Multiaddr) transport.Transport {
 	s.transports.RLock()
 	defer s.transports.RUnlock()
 	if len(s.transports.m) == 0 {
-		log.Error("you have no transports configured")
+		// make sure we're not just shutting down.
+		if s.transports.m != nil {
+			log.Error("you have no transports configured")
+		}
 		return nil
 	}
 
@@ -77,6 +83,9 @@ func (s *Swarm) AddTransport(t transport.Transport) error {
 
 	s.transports.Lock()
 	defer s.transports.Unlock()
+	if s.transports.m == nil {
+		return ErrSwarmClosed
+	}
 	var registered []string
 	for _, p := range protocols {
 		if _, ok := s.transports.m[p]; ok {
--- a/transport_test.go
+++ b/transport_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	swarm "github.com/libp2p/go-libp2p-swarm"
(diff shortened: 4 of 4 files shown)


vyzo (approved):

swarm.go
```diff
@@ -176,6 +177,20 @@ func (s *Swarm) teardown() error {
...
 
+	// Now close out any transports (if necessary).
+	s.transports.Lock()
+	transports := s.transports.m
+	s.transports.m = nil
+	s.transports.Unlock()
+
+	for t := range transports {
+		if closer, ok := t.(io.Closer); ok {
+			if err := closer.Close(); err != nil {
```
Jorropo, line 188: In my case (tor and i2p) `closer.Close()` is slow, it needs to call other peers in the dht about him leaving, that takes some times, I would like each transport to have his own goroutine.

Jorropo (changes requested):

Jorropo (approved):

Jorropo (approved):

swarm.go
Stebalien: Eh, I'd prefer to be able to _wait_ and I'm not too concerned about the cost of firing off goroutines on shutdown.

vyzo (approved):

Jorropo (approved):
LGTM I couldn't ask for more 👌

Jorropo (changes requested):

swarm.go
```diff
@@ -176,6 +177,27 @@ func (s *Swarm) teardown() error {
...
 	s.refs.Wait()
 
+	// Now close out any transports (if necessary). Do this after closing
+	// all connections/listeners.
+	s.transports.Lock()
+	transports := s.transports.m
+	s.transports.m = nil
+	s.transports.Unlock()
+
+	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
Stebalien: That isn't safe if `Close` is called concurrently. Basically, we can only call `Add` on this waitgroup when we can guarantee that the waitgoup won't be zero.

swarm.go
```diff
@@ -176,6 +177,27 @@ func (s *Swarm) teardown() error {
 	// Wait for everything to finish.
 	s.refs.Wait()
 
+	// Now close out any transports (if necessary). Do this after closing
```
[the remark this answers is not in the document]
Stebalien: I'm not going to take a lock in one function, and release it in another. The correct fix here would be to call `refs.Add(1)` *before* starting a dial, releasing the ref on dial failure.

jacobheun:
What's the state of this, is it blocked by anything?

vyzo:
not that I know, LGTM. @Stebalien ?

Stebalien:
I haven't merged it because it's still technically possible to call `Dial` on transports after calling `Close`. Honestly, merging with that issue is probably fine for now.

Jorropo (approved):
I've changed my mind, that a close, as long as it works performance is not an issue.

Jorropo:
@Stebalien what is the status of this ?
I need it for https://github.com/berty/go-libp2p-tor-transport, without it this is causing memory, cpu and process leakage.
Plus I think there is no need to bother as much as I were doing, it's fine for me.

Stebalien:
Yeah, this is probably fine as-is. I'll merge for now with an open issue: https://github.com/libp2p/go-libp2p-swarm/issues/230.

Stebalien:
Actually, I do need a final signoff from @raulk on this.

swarm.go
```diff
@@ -176,6 +177,27 @@ func (s *Swarm) teardown() error {
...
+	s.transports.m = nil
+	s.transports.Unlock()
+
+	var wg sync.WaitGroup
+	for _, t := range transports {
+		if closer, ok := t.(io.Closer); ok {
+			wg.Add(1)
+			go func(c io.Closer) {
+				defer wg.Done()
+				if err := closer.Close(); err != nil {
```
raulk, line 193: A log statement confirming that we're closing down this closeable transport would be nice. But then again, we need better logging all throughout, so I won't object if we don't add this one here right now.

swarm_listen.go
```diff
@@ -40,7 +40,12 @@ func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
 func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
 	tpt := s.TransportForListening(a)
 	if tpt == nil {
-		return ErrNoTransport
+		select {
+		case <-s.proc.Closing():
+			return ErrSwarmClosed
```
raulk, line 45: I'm not sure I understand why this here? Which edge case is this addressing? Either we accept or we don't accept adding listen addresses while the swarm is closing/closed, more generally, but this looks weird?

swarm_transport.go
```diff
@@ -48,7 +51,10 @@ func (s *Swarm) TransportForListening(a ma.Multiaddr) transport.Transport {
 	s.transports.RLock()
 	defer s.transports.RUnlock()
 	if len(s.transports.m) == 0 {
-		log.Error("you have no transports configured")
+		// make sure we're not just shutting down.
+		if s.transports.m != nil {
```
raulk, line 55: Would be nice to actually model a swarm status with atomic ints. Inferring we're closing down based on nil vs. 0-length slice differentiation is a bit of a footgun.

raulk (commented):
I'm ok to merge these changes, but there are some edge cases that I quite don't understand. And the swarm needs to store its state (unstarted, started, closing, closed) in an atomic int so we can implement a state machine, rather than rely on hacks to infer which state we're in.

swarm_transport.go
Stebalien: Note: this is a map, not a slice. We could have a second flag, but I'd prefer to avoid storing the state twice.

swarm_listen.go
Stebalien: I'll add a comment: TL;DR: we could fail to get the transport either because we're closed, or because the transport was never registered.

Stebalien:
Tests are broken due to #246. Otherwise, they passed.

PR #229, opened 2020-09-02
fix: handle case where swarm closes before stream

Stebalien:
When we close a connection, we set the "stream" set to nil to avoid opening new stream. Unfortunately, this meant we wouldn't decrement the reference count on the swarm.


vyzo (approved):
good catch!

PR #233, opened 2021-01-14
Support for Hole punching

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/1039.

To implement the hole punching co-ordination as described in  https://github.com/libp2p/specs/pull/173, we need to 
support forcing a direct connection (if possible) even if a Relayed connection to a peer already exists.

### TODO
- [ ] Update & propagate deps.


swarm.go
```diff
@@ -395,15 +397,32 @@ func (s *Swarm) bestConnToPeer(p peer.ID) *Conn {
...
+				best = c
+				bestLen = cLen
+			}
+		} else {
+			// Since the best connection we have till now is a Relayed connection,
+			// we will simply replace with a newer connection if it has atleast as many streams.
+			if cLen >= bestLen {
+				best = c
+				bestLen = cLen
+			}
```
marten-seemann, line 416: Probably easier:
```go
if isDirectConn(best) && !isDirectConn(c) {
    continue
}
if cLen >= bestLen {
     best = c
     bestLen = cLen
}
```

swarm.go
```diff
@@ -395,15 +397,32 @@ func (s *Swarm) bestConnToPeer(p peer.ID) *Conn {
...
 		}
-
 	}
 	return best
 }
 
+func isDirectConn(c *Conn) bool {
+	return c != nil && !c.conn.Transport().Proxy()
```
marten-seemann, line 423: Can `c == nil` ever happen here?

swarm_dial.go
```diff
@@ -251,9 +251,14 @@ func (s *Swarm) dialPeer(ctx context.Context, p peer.ID) (*Conn, error) {
 
 	defer log.EventBegin(ctx, "swarmDialAttemptSync", p).Done()
 
-	// check if we already have an open connection first
 	conn := s.bestConnToPeer(p)
-	if conn != nil {
+	forceDirect, _ := network.GetForceDirectDial(ctx)
+	if forceDirect {
```
marten-seemann, line 256: Nit: this could be a one-line `if`.

marten-seemann (commented):
Why do we use the `context` to store if we want to force a direct dial? This is not the most obvious API, took me a minute to realize what's going on here. Is there a better solution?
If not, can we add an explanation to `Swarm.dial` why this route was chosen?

aarshkshah1992:
cc @vyzo

swarm.go
aarshkshah1992: Yeah, if no connection exists.

swarm.go
```diff
@@ -378,9 +379,10 @@ func (s *Swarm) ConnsToPeer(p peer.ID) []network.Conn {
 
 // bestConnToPeer returns the best connection to peer.
 func (s *Swarm) bestConnToPeer(p peer.ID) *Conn {
+
 	// Selects the best connection we have to the peer.
-	// TODO: Prefer some transports over others. Currently, we just select
-	// the newest non-closed connection with the most streams.
+	// Prefers direct connections over Relayed connections.
+	// For tie-breaking, select the newest non-closed connection with the most streams.
```
vyzo, line 385: let's keep this TODO around? We do want to prioritize transports eventually.

swarm.go
vyzo, line 416: Actually this is not correct: we always want to replace a relayed connection with a direct connection, regardless of how many streams it has.

swarm_dial.go
```diff
@@ -383,20 +403,28 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		return nil, &DialError{Peer: p, Cause: ErrNoAddresses}
 	}
 	goodAddrs := s.filterKnownUndialables(p, peerAddrs)
+	// TODO Consider dialing private addresses as well
+	// See https://github.com/libp2p/specs/pull/173/files#r288784559
+	if forceDirect {
+		goodAddrs = addrutil.FilterAddrs(goodAddrs, manet.IsPublicAddr, s.nonProxyAddr)
+	}
```
vyzo, line 410: hrm... yeah, this could be problematic here. If we are behind the same NAT we want the private dial. 
Let's not filter only public addrs?

vyzo (commented):
A couple of small but important points that should be addressed.

swarm.go
aarshkshah1992: Ah, missed this. Have fixed it.

aarshkshah1992: Have implemented something similar now.

aarshkshah1992 (commented):
@vyzo  Have addressed your review. Once you approve, will propagate released deps and merge.

swarm_dial.go
```diff
@@ -516,7 +542,6 @@ dialLoop:
 				remoteAddrChan = nil
 				continue
 			}
-
```
vyzo, line 519: restore this line pls.

swarm_dial.go
```diff
@@ -383,20 +403,26 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
 		return nil, &DialError{Peer: p, Cause: ErrNoGoodAddresses}
 	}
 
-	/////// Check backoff andnRank addresses
-	var nonBackoff bool
-	for _, a := range goodAddrs {
-		// skip addresses in back-off
-		if !s.backf.Backoff(p, a) {
-			nonBackoff = true
+	// TODO How do we do backoff for forcedirect
```
vyzo, line 413: this needs to be addressed.

swarm_dial.go
```diff
@@ -301,15 +311,19 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+			}
+		} else if conn != nil {
 			// Hm? What error?
 			// Could have canceled the dial because we received a
 			// connection or some other random reason.
 			// Just ignore the error and return the connection.
 			log.Debugf("ignoring dial error because we have a connection: %s", err)
 			return conn, nil
 		}
-
```
vyzo, line 312: restore this line pls.

swarm_dial.go
```diff
@@ -301,15 +311,19 @@ func (s *Swarm) doDial(ctx context.Context, p peer.ID) (*Conn, error) {
 	conn, err := s.dial(ctx, p)
 	if err != nil {
 		conn = s.bestConnToPeer(p)
-		if conn != nil {
+		if forceDirect {
+			if isDirectConn(conn) {
+				log.Debugf("ignoring dial error because we have a connection: %s", err)
```
vyzo, line 316: this log could be better, perhaps ".. because we already have a direct connection ..."?

vyzo (commented):
looks good overall, a few stylistic comments.

One issue that remains is what do with the backoff -- see the TODO. Shouldn't this be addressed before merge?

swarm_dial.go
aarshkshah1992: @vyzo 

Honestly, I think we should just NOT do backoffs for forced direct conns at all. We only do this during hole punching anyways and a hole punching attempt can fail for a variety of reasons other than the address being bad.

@Stebalien Any thoughts ?

vyzo: Ok, fair enough, I am fine with it.

aarshkshah1992: I've removed this TODO.

swarm_dial.go
aarshkshah1992: done.

swarm_dial.go
aarshkshah1992: done.

aarshkshah1992:
@vyzo Green tick please !!!!!!!!!!!!!!!!!!!

swarm.go
```diff
@@ -392,15 +393,17 @@ func (s *Swarm) ConnsToPeer(p peer.ID) []network.Conn {
...
+	// TODO: Prefer some transports over others.
+	// For now, prefers direct connections over Relayed connections.
+	// For tie-breaking, select the newest non-closed connection with the most streams.
 	s.conns.RLock()
 	defer s.conns.RUnlock()
 
 	var best *Conn
 	bestLen := 0
-	for _, c := range s.conns.m[p] {
+	for i := range s.conns.m[p] {
```
vyzo, line 405: so we don't use the `i` anywhere, why not write this as `for _, c := ...` as it was before?

vyzo (commented):
one final nit.

swarm.go
aarshkshah1992: Let me know if I'm wrong but for a slice of reference types(pointers), the reference semantic form of the for loop works better than the value semantic form ?

swarm_dial.go
aarshkshah1992: done.

swarm.go
aarshkshah1992: Given that we are looping over a collection of reference types(pointers), it's easier to reason about the code if we use the reference semantic form of the range loop and avoid copies:

https://www.ardanlabs.com/blog/2017/06/for-range-semantics.html

vyzo: this is nonsense; we don't use the i and we iterate over connections. 
We should use the builtin syntax for that, which turns one line into two.

aarshkshah1992: done.

vyzo (approved):

dial_sync.go
```diff
@@ -96,13 +96,13 @@ func (ad *activeDial) start(ctx context.Context) {
...
 }
 
-func (ds *DialSync) getActiveDial(p peer.ID) *activeDial {
+func (ds *DialSync) getActiveDial(ctx context.Context, p peer.ID) *activeDial {
 	ds.dialsLk.Lock()
 	defer ds.dialsLk.Unlock()
 
 	actd, ok := ds.dials[p]
 	if !ok {
-		adctx, cancel := context.WithCancel(context.Background())
```
Stebalien, line 105: This _needs_ to use the background context. Otherwise, if the first dialer cancels, we'll cancel the overall dial.

swarm_dial.go
```diff
@@ -321,6 +336,11 @@ func (s *Swarm) canDial(addr ma.Multiaddr) bool {
 	return t != nil && t.CanDial(addr)
 }
 
+func (s *Swarm) nonProxyAddr(addr ma.Multiaddr) bool {
+	t := s.TransportForDialing(addr)
+	return !t.Proxy()
```
Stebalien, line 341: This could panic. I know we don't call it in a way that can panic, but it could still panic and someone else could call it.

dial_sync.go
Stebalien: This is also why the tests are failing.

PR #235, opened 2021-01-28
Enable QUIC in Test Swarm

aarshkshah1992:
Makes it easy for tests Hosts across the board to have a QUIC transport.


willscott (approved):

Stebalien (approved):
We really nee to extract this package somewhere else, but this works for now.

PR #236, opened 2021-02-03
Treat transient connections as opt-in when opening new streams

vyzo:
This adds support for transient connections with opt-in semantics; when the best connection to a peer is transient, then only open a stream if the user is prepared to deal with it and has specified the relevant context option.

Rationale: many protocols may not be prepared to deal with transient connections; this makes it explicit opt-in to protect them and also eliminate interference with connection upgrade attempts.

depends on https://github.com/libp2p/go-libp2p-core/pull/175

### TODO
- [x] Release & update deps.


aarshkshah1992:
@vyzo 

We still need the code to mark limited Relay connections as Transient. How and where in the code do you plan to do that without leaking the idea of limited Relays to the Swarm ?

vyzo:
The relay dialer/listener constructs the Conn object, so it can mark itself when it knows it's transient.
No need to leak into the swarm code at all!

vyzo:
retract that, the swarm creates the stat object -- we need a mechanism to pass stuff from the transport all the way up to the swarm.

vyzo:
added threading of stat if available from the transport conn.

swarm.go
```diff
@@ -351,6 +358,13 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
 				return nil, err
 			}
 		}
+
+		if c.Stat().Transient {
```
aarshkshah1992, line 362: can you add a unit test for this ?

aarshkshah1992 (commented):
Lgtm. Just needs a unit test.

aarshkshah1992 (approved):

swarm.go
vyzo: it would be nice, but this is quite hard to write and probably not worth the effort.

aarshkshah1992:
@vyzo 
 
Should I go ahead and merge this ?

vyzo:
waiting for travis, I will merge and cut release.

PR #237, opened 2021-02-08
prevent dialing addresses that we're listening on

marten-seemann:
It's impossible to run two nodes on the same IP:port, so we know for sure that any dial to an address that we're listening on will fail (as the peer IDs won't match).
In practice, this will be most useful for preventing dials to localhost for nodes that are listening on the default port.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -440,7 +440,7 @@ func (s *Swarm) filterKnownUndialables(p peer.ID, addrs []ma.Multiaddr) []ma.Mul
 	for _, addr := range lisAddrs {
 		protos := addr.Protocols()
 		// we're only sure about filtering out /ip4 and /ip6 addresses, so far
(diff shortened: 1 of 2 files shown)


vyzo (commented):
hopefully this wont break relay dialing; it shouldn't because it shouldn't show in interface addrs, but mind checking?

willscott:
can we reuse the [`filterKnownUndialables`](https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm_dial.go#L437)  logic?

Stebalien (changes requested):
We're already doing this on line 449. I think the issue is that we're _only_ filtering IP+TCP and need to adjust  line 443 to allow any address starting with an IP addr (not just length 2 ones).

marten-seemann:
@Stebalien You're right. I removed the check for length 2, as QUIC returns 3 protocols here: IPv{4,6}, UDP, QUIC.

Stebalien (approved):

PR #238, opened 2021-02-09
only listen on localhost in tests

marten-seemann:
OSX complains when you listen on all interfaces:
Screen Shot 2021-02-09 at 11 09 34


Stebalien (approved):
This _should_ work. And if it doesn't, we need to fix the tests that fail anyways.

PR #239, opened 2021-02-12
avoid assigning a function to a variable

marten-seemann:
No functional change expected. I believe defining the function on the struct itself is more efficient than assigning it to a variable.
Really, a lambda should only be used to capture a variable, which is not done here.


Stebalien (approved):
I assume go is decent at optimizing this, but you're right. This is cleaner.

PR #241, opened 2021-02-19
Fix swarm transient connections logic

aarshkshah1992:
@vyzo Even the `conn.NewStream` function is exported and called in `Host`.


vyzo (approved):
ah, I hadnt realized this is exported; I guess it's used from identify.

PR #246, opened 2021-03-29
Fix simultaneous dials

vyzo:
This implements the necessary logic for dealing with simultaneous dials which may have conflicting constraints.

TBD: tests


dial_sync.go
```diff
@@ -34,17 +41,61 @@ type activeDial struct {
...
+	donech chan struct{}
 
 	ds *DialSync
 }
 
-func (ad *activeDial) wait(ctx context.Context) (*Conn, error) {
+func (ad *activeDial) dial(ctx context.Context) (*Conn, error) {
 	defer ad.decref()
+
+	dialCtx := ad.dialContext(ctx)
```
Stebalien, line 64: nit: by go convention, `xContext` means ["x but with a context"](https://golang.org/pkg/net/#Dialer.DialContext). Rename to `deriveDialContext`.

dial_sync.go
```diff
@@ -127,14 +229,27 @@ func (ds *DialSync) getActiveDial(p peer.ID) *activeDial {
...
+	for {
+		ad = ds.getActiveDial(p)
+
+		// signal the start of dial
+		select {
+		case ad.dialch <- struct{}{}:
+			break startDial
+		case <-ad.waitch:
+			// we lost a race, we need to try again because the connection might not be what we want
+			ad.decref()
```
Stebalien, line 244: Why are we dereferencing? Won't that cause the in-progress dial to be canceled if the original caller bails?

Stebalien: I mean, we'll just restart, but that's pretty inefficient if we were half-way through opening a bunch of connections.

swarm_conn.go
```diff
@@ -54,6 +54,11 @@ func (c *Conn) ID() string {
 // open notifications must finish before we can fire off the close
 // notifications).
 func (c *Conn) Close() error {
+	// for mock tests
+	if c == nil || c.conn == nil {
+		return nil
+	}
+
```
Stebalien, line 61: ???? This is really scary. Why?

dial_sync.go
```diff
@@ -53,47 +104,96 @@ func (ad *activeDial) wait(ctx context.Context) (*Conn, error) {
...
+	}
+
+	return dialCtx
+}
+
+func (ad *activeDial) dedup(addrs []ma.Multiaddr) (result []ma.Multiaddr) {
+	ad.addrsLk.Lock()
+	defer ad.addrsLk.Unlock()
+
+	for _, a := range addrs {
```
Stebalien, line 127: What if we never bother to dial one of these addresses because we're canceled? We'll just skip that overall.

Test case:

1. Start dial A.
2. Start dial B.
3. Cancel dial A.
4. Dial B should pick up where dial A left off.

With the current code, dial B will, I believe, fail to dial any of the addresses dial A was trying to dial.

Stebalien (changes requested):
How about an approach where a new dial joins existing dials?

1. Spawn a dial worker.
2. Each dial sends a message down a channel to the dial worker with options & address filters.
3. When a dial joins, check all undialed addresses (addresses in the peerstore minus the set of already tried addresses) against the address filter and consider adding them to the back of the dial queue.
3. When we form connections, check them against the address filters.
4. When a dial "leaves" (cancels), remove addresses that no longer match any filter.

I'm open to approaches, but the important parts are:

1. Seamlessly handle multiple dials joining/leaving without wasting work.
2. Eventually, we want this to morph into something like the "dialer v2 work". Whatever we do here, we want to be able to "replan" dials to pending addresses while the dial is active. The current approach will force us to wait for the current dial to complete.

swarm_conn.go
vyzo: unnecessary after fixing a bug, will remove!

dial_sync.go
vyzo: no, the dials will still happen -- they use the background context, so there should be no interference.

Stebalien: Ah, I see.

dial_sync.go
Stebalien: Turns out, we only invoke this code if the active dial is currently "finishing". So we're not going to cancel anything.

vyzo:
Closing in favor of #250 which accomplishes the same goals with much cleaner code.

PR #247, opened 2021-03-29
fix: use 64bit stream/conn IDs

Stebalien:
Given 1k requests per second (assuming one stream per request), we can easily loop around the stream ID after less than 2 months. 32bits is plenty (usually) for connection-scoped stream IDs because individual connections don't usually last that long, but isn't enough for a _global_ stream ID.

Given that there's no reason for these to be 32bit IDs, let's just make them 64bits.


vyzo (approved):

Stebalien:
(test failure is known and being fixed)

PR #252, opened 2021-04-01
Dial batching

vyzo:
Extracted from #250 

This adds dial batching logic to group together dials while leaving space for new dials to come in and join.
The initial logic simply batches together dials, and moves to the next batch with a context switch and no delay.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -480,17 +480,37 @@ loop:
 			}
 
 		case <-triggerDial:
-			for _, addr := range nextDial {
+			// we dial batches of addresses together, logically belonging to the same batch
+			// after a batch of addresses has been dialed, we add a delay before initiating the next batch
+			dialed := false
+			last := 0
+			next := 0
+			for i, addr := range nextDial {
+				if dialed && !s.sameAddrBatch(nextDial[last], addr) {
+					break
+				}
+
+				next = i + 1
+
 				// spawn the dial
 				ad := pending[addr]
 				err := s.dialNextAddr(ad.ctx, p, addr, resch)
 				if err != nil {
 					dispatchError(ad, err)
+					continue
 				}
+
+				ad.dialed = true
+				dialed = true
+				last = i
+				active++
 			}
 
-			nextDial = nil
-			triggerDial = nil
+			nextDial = nextDial[next:]
+			if !dialed || len(nextDial) == 0 {
+				// we didn't dial anything because of backoff or we don't have any more addresses
+				triggerDial = nil
+			}
 
 		case res := <-resch:
 			active--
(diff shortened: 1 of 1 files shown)


swarm_dial.go
```diff
@@ -480,17 +480,37 @@ loop:
...
+				ad.dialed = true
+				dialed = true
+				last = i
+				active++
 			}
 
-			nextDial = nil
-			triggerDial = nil
+			nextDial = nextDial[next:]
+			if !dialed || len(nextDial) == 0 {
```
Stebalien, line 510: if `dialed` is false, won't `len(nextDial)` always be 0?

swarm_dial.go
```diff
@@ -480,17 +480,37 @@ loop:
...
 
 		case <-triggerDial:
-			for _, addr := range nextDial {
+			// we dial batches of addresses together, logically belonging to the same batch
+			// after a batch of addresses has been dialed, we add a delay before initiating the next batch
+			dialed := false
+			last := 0
+			next := 0
+			for i, addr := range nextDial {
+				if dialed && !s.sameAddrBatch(nextDial[last], addr) {
```
Stebalien, line 489: assuming we actually had delays, wouldn't any new addresses trigger the next group? I think we need to add delays so we can at least reason about this.

Stebalien: Assuming we fixed that, I think we'd have a new problem: _better_ dials (e.g., hole punching) might get stuck behind an ongoing "worse" dial group.

Stebalien (commented):
I _think_ we may need something more like dial "lanes". We already put addresses into lanes (when we rank them) but then we concatenate these lanes together.

We could actually just keep the lanes and keep an index of the "current" lane.

1. Whenever we trigger a dial (no active dials, timeout, etc.) we'd dial everything through the current lane +1.
2. Whenever a joining dial adds new addresses, we immediately spawn dials for all new addresses in the lanes we're currently dialing.

Thoughts on this?

swarm_dial.go
vyzo: yes, but we could have dialed=true and no more addresses, so we need to check nextDial as well.

swarm_dial.go
vyzo: Not necessarily.

When new addresses come in, we append to existing pending dials and _rerank_ them, so they would get pushed at the top.
At this point we could just trigger a new dial to get them processed.

vyzo:
> We could actually just keep the lanes and keep an index of the "current" lane.
>
>  1. Whenever we trigger a dial (no active dials, timeout, etc.) we'd dial everything through the current lane +1.
>   2.  Whenever a joining dial adds new addresses, we immediately spawn dials for all new addresses in the lanes we're currently dialing.

I like this idea, although we could also do it with a flat address slice and the `sameAddrBatch` checking for lane crosses.
Regardless, we should definitely trigger new dials whenever we add new addrs in a lane lower than the last dial.

swarm_dial.go
Stebalien: Ah, I see.

Stebalien: Ah, wait, no. My point is that we can just get rid of the `dialed` variable, right?

vyzo: I guess we could, but it would be useful when/if we introduce delays so that we know that `last` is valid.

Stebalien:
I think we'll need actual delays to move this forward. Otherwise, it just adds complexity without improving anything.

vyzo:
agreed! I'll add the relevant logic.

Stebalien:
Closing because we'll need another pass of design and likely won't get to this for a while.

PR #254, opened 2021-04-15
Feat/fix staticcheck

coryschwartz:
Context: https://github.com/protocol/.github/issues/41


dial_test.go
```diff
@@ -437,12 +437,13 @@ func TestDialBackoffClears(t *testing.T) {
...
 	before := time.Now()
-	if c, err := s1.DialPeer(ctx, s2.LocalPeer()); err == nil {
+	c, err := s1.DialPeer(ctx, s2.LocalPeer())
+	if err == nil {
 		t.Fatal("dialing to broken addr worked...", err)
-		defer c.Close()
 	} else {
 		t.Log("correctly got error:", err)
 	}
+	defer c.Close()
```
marten-seemann, line 446: If `err != nil`, then `c` might be `nil`. The `defer` should be moved before the `t.Fatal`.

swarm_dial.go
```diff
@@ -237,18 +237,18 @@ func (s *Swarm) DialPeer(ctx context.Context, p peer.ID) (network.Conn, error) {
...
-	var logdial = lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
+	lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
 	err := p.Validate()
 	if err != nil {
 		return nil, err
 	}
 
 	if p == s.local {
-		log.Event(ctx, "swarmDialSelf", logdial)
+		// log.Event(ctx, "swarmDialSelf", logdial)
```
marten-seemann, line 247: @Stebalien, how are we supposed to deal with this?

swarm_dial.go
```diff
@@ -237,18 +237,18 @@ func (s *Swarm) DialPeer(ctx context.Context, p peer.ID) (network.Conn, error) {
 // dialbackoff.
 func (s *Swarm) dialPeer(ctx context.Context, p peer.ID) (*Conn, error) {
 	log.Debugf("[%s] swarm dialing peer [%s]", s.local, p)
-	var logdial = lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
+	lgbl.Dial("swarm", s.LocalPeer(), p, nil, nil)
```
Stebalien, line 240: This doesn't actually do anything. We should replace these two lines with:

```go
log.Debugw("dialing peer", "from", s.local, "to", p)
```

swarm_dial.go
```diff
@@ -237,18 +237,18 @@ func (s *Swarm) DialPeer(ctx context.Context, p peer.ID) (network.Conn, error) {
...
 	}
 
 	if p == s.local {
-		log.Event(ctx, "swarmDialSelf", logdial)
+		// log.Event(ctx, "swarmDialSelf", logdial)
 		return nil, ErrDialToSelf
 	}
 
-	defer log.EventBegin(ctx, "swarmDialAttemptSync", p).Done()
+	// defer log.EventBegin(ctx, "swarmDialAttemptSync", p).Done()
```
Stebalien, line 251: Let's just delete these lines.

swarm_listen.go
```diff
@@ -111,7 +111,7 @@ func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
 					// ignore.
 					return
 				default:
-					log.Warningf("add conn %s failed: ", err)
+					log.Warnf("add conn %s failed: ", err)
```
Stebalien, line 114: If we're changing these, we might as well switch to a structured version:

```go
log.Warnw("adding connection failed", "to", p, "error", err)
```

Stebalien (changes requested):

dial_test.go
marten-seemann: Fixed in #255.

swarm_dial.go
marten-seemann: Fixed in #255.

swarm_listen.go
marten-seemann: Fixed in #255.

swarm_dial.go
Stebalien: `log.Debugw` with a message.

PR #257, opened 2021-04-20
fix: avoid returning typed nils

Stebalien:


marten-seemann (approved):
It doesn't hurt, but I don't think there's a rule that says "If `err != nil`, the other return values must be `nil` / their default value". I had assumed the convention is that if `err != nil`, you can't make any assumption about the other return values.

Stebalien:
Yeah, it's just confusing to have typed nils.

PR #260, opened 2021-05-18
remove incorrect call to InterceptAddrDial

marten-seemann:
`addConn` is called both when we add a dialed and an accepted connection to the swarm. `InterceptAddrDial` is only supposed to intercept outgoing connections though. When dialing, we already call `InterceptAddrDial` when we compose the
list of dialable addresses: https://github.com/libp2p/go-libp2p-swarm/blob/eef02f818639129a7184dba4ac0dbba1471a4b40/swarm_dial.go#L658-L678

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -212,16 +212,6 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 		addr = tc.RemoteMultiaddr()
 	)
 
-	if s.gater != nil {
-		if allow := s.gater.InterceptAddrDial(p, addr); !allow {
-			err := tc.Close()
-			if err != nil {
-				log.Warnf("failed to close connection with peer %s and addr %s; err: %s", p.Pretty(), addr, err)
-			}
-			return nil, ErrAddrFiltered
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


raulk (commented):
This looks fair, and intercepting outbound dials is what the contract says, so if users are using this to intercept inbounds, it's wrong usage so no need to do change management.

However, we should add an integration test here swarm <=> conngater. It makes me uneasy that this was uncaught and there are no tests here.

Stebalien (approved):

prestonvanloon:
Hey guys, we have been reviewing recent updates to libp2p for https://github.com/prysmaticlabs/prysm/pull/9772 and this PR seems to modify expected behavior of the ConnectionGater. 

Please correct me if I am misunderstanding this change. Does this change modify the behavior such that a ConnectionGater will no longer actively reject inbound and outbound connections? We are following the description in the upstream godoc here: https://github.com/libp2p/go-libp2p/blob/v0.15.1/options.go#L329-L333

marten-seemann:
`InterceptAddrDial` was (incorrectly) called for outgoing connections. This PR fixes that.

PR #268, opened 2021-08-22
remove unused context in Swarm.dialWorkerLoop

marten-seemann:


Stebalien (approved):
Ah, so we just wait for the request channel to terminate? Makes sense.

marten-seemann:
> Ah, so we just wait for the request channel to terminate? Makes sense.

Apparently that's what we do. Just trying to clean things up a bit.
I honestly don't understand the 250 line `dialWorkerLoop`. There's not a lot of useful documentation in the code, and I don't find it very obvious to begin with.

PR #271, opened 2021-08-23
update badges

marten-seemann:


Stebalien (approved):
We probably need to do a massive sed.

PR #272, opened 2021-08-23
simplify the DialSync code

marten-seemann:
It's easier to reason about this code if `activeDial` doesn't contain a pointer back to `DialSync` (which already has a map of `activeDial`s). It also allows us to remove the memory footprint of the `activeDial` struct, so this should be (slightly) more efficient.

No functional change expected.


Stebalien:
Flagging others for review. We need more people who understand this code.

Stebalien (approved):

PR #277, opened 2021-09-05
read and use the direction from the simultaneous connect context

marten-seemann:
This depends on go-libp2p-core with https://github.com/libp2p/go-libp2p-core/pull/210 merged. Will rebase this PR once v0.10.0 is out.


dial_sync.go
```diff
@@ -56,8 +56,8 @@ func (ad *activeDial) dial(ctx context.Context, p peer.ID) (*Conn, error) {
 	if forceDirect, reason := network.GetForceDirectDial(ctx); forceDirect {
 		dialCtx = network.WithForceDirectDial(dialCtx, reason)
 	}
-	if simConnect, reason := network.GetSimultaneousConnect(ctx); simConnect {
-		dialCtx = network.WithSimultaneousConnect(dialCtx, reason)
+	if simConnect, isClient, reason := network.GetSimultaneousConnect(ctx); simConnect {
+		dialCtx = network.WithSimultaneousConnect(dialCtx, isClient, reason)
```
Stebalien, line 60: We really should have a test for this.

Stebalien (approved):

dial_sync.go
marten-seemann: Yeah, but this has to wait until we have the `SecureTransport` / `SecureMuxer` changes in place.

marten-seemann:
go-libp2p-core v0.10.0 has multiple breaking changes, so we can't just use that version without also fixing the other changes. I'll keep this PR as is. Note that it references go-libp2p-core@094b0d3f8ba2, which is a commit on master, so that import path should be safe from garbage collection.

PR #278, opened 2021-09-05
stop using goprocess to control teardown

marten-seemann:
Depends on https://github.com/libp2p/go-libp2p-core/pull/212.

This PR:
* removes `Swarm.Process() goprocess.Process`
* removes `Swarm.Context() context.Context`
* removes the `context.Context` from `NewSwarm`

Annoyingly, this means that we can't attach the closing of the peerstore in the `testing` package any more:
https://github.com/libp2p/go-libp2p-swarm/blob/29bd937b3098b5b8a9d74a7b393d5ea1ed1fdd10/testing/testing.go#L118-L120
We therefore return a `network.Network` from `testing.GenSwarm`, and close the peerstore in `Close`. This change is responsible for 95% of the LOC changed in this PR.


swarm.go
```diff
@@ -201,11 +191,6 @@ func (s *Swarm) teardown() error {
 	return nil
 }
 
-// Process returns the Process of the swarm
-func (s *Swarm) Process() goprocess.Process {
-	return s.proc
```
Stebalien, line 206: For posterity: this was very difficult to use correctly because _adding_ child processes raced with the process shutting down (leading to a panic if _this_ process closed before the child was added).

Technically, it could be used correctly with `AddChildNoWait`, but most callers would use `AddCHild` and `Go` (leading to panics on shutdown).

testing/testing.go
```diff
@@ -88,8 +86,18 @@ func GenUpgrader(n *swarm.Swarm) *tptu.Upgrader {
 	}
 }
 
+type mSwarm struct {
```
Stebalien, line 89: Note: It's probably safe to use `testing.Cleanup` to close the peerstore.

transport_test.go
```diff
@@ -42,24 +46,23 @@ func (dt *dummyTransport) Close() error {
 	return nil
 }
 
+type swarmWithTransport interface {
```
Stebalien, line 49: We have this as `network.TransportNetwork`.

testing/testing.go
```diff
@@ -88,8 +86,18 @@ func GenUpgrader(n *swarm.Swarm) *tptu.Upgrader {
...
+}
+
+func (s *mSwarm) Close() error {
+	s.ps.Close()
+	return s.Swarm.Close()
+}
+
 // GenSwarm generates a new test swarm.
-func GenSwarm(t *testing.T, ctx context.Context, opts ...Option) *swarm.Swarm {
+func GenSwarm(t *testing.T, opts ...Option) network.Network {
```
Stebalien, line 100: This change will likely affect other modules so it would be nice if we could avoid it (see my comment above about using `testing.Cleanup`.

swarm.go
```diff
@@ -103,11 +102,14 @@ type Swarm struct {
 // `extra` interface{} parameter facilitates the future migration. Supported
 // elements are:
 //  - connmgr.ConnectionGater
-func NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, extra ...interface{}) *Swarm {
+func NewSwarm(local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, extra ...interface{}) *Swarm {
```
Stebalien, line 105: If we're breaking this interface anyways, let's consider using functional options. We can also do that in a followup patch.

Stebalien (changes requested):
:heart:

testing/testing.go
marten-seemann: Oh cool, I wasn't aware of `testing.Cleanup`. That's exactly what I've been looking for!

Stebalien: I think it's relatively new.

swarm.go
```diff
@@ -124,23 +126,11 @@ func NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc
...
-	s.proc.SetTeardown(s.teardown)
-
 	return s
 }
 
-func (s *Swarm) teardown() error {
-	// Wait for the context to be canceled.
-	// This allows other parts of the swarm to detect that we're shutting
-	// down.
-	<-s.ctx.Done()
```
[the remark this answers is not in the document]
marten-seemann: Yes. Added it to `Close()`.

swarm.go
```diff
@@ -124,23 +126,11 @@ func NewSwarm(ctx context.Context, local peer.ID, peers peerstore.Peerstore, bwc
...
-
+func (s *Swarm) Close() error {
```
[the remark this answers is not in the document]
marten-seemann: I like to use a `sync.Once` for that, as it blocks all calls until the first one is done.

Stebalien (approved):

marten-seemann:
Deferring merging until we've cut a new -core release.

PR #279, opened 2021-09-07
use functional options to configure the swarm

marten-seemann:
As suggested in https://github.com/libp2p/go-libp2p-swarm/pull/278#discussion_r703342336.
This PR depends on #278.

When reviewing, please just review the most recent commit.

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -50,6 +50,13 @@ func WithConnectionGater(gater connmgr.ConnectionGater) Option {
 	}
 }
 
+// WithMetrics sets a metrics reporter
+func WithMetrics(reporter metrics.Reporter) Option {
+	return func(s *Swarm) {
+		s.bwc = reporter
+	}
+}
+
 // Swarm is a connection muxer, allowing connections to other peers to
 // be opened and closed, while still using the same Chan for all
 // communication. The Chan sends/receives Messages, which note the
@@ -107,12 +114,11 @@ type Swarm struct {
 }
 
 // NewSwarm constructs a Swarm.
-func NewSwarm(local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, opts ...Option) *Swarm {
+func NewSwarm(local peer.ID, peers peerstore.Peerstore, opts ...Option) *Swarm {
 	ctx, cancel := context.WithCancel(context.Background())
 	s := &Swarm{
 		local:     local,
 		peers:     peers,
-		bwc:       bwc,
 		ctx:       ctx,
 		ctxCancel: cancel,
 	}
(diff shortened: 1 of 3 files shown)


swarm.go
```diff
@@ -98,12 +107,7 @@ type Swarm struct {
 }
 
 // NewSwarm constructs a Swarm.
-//
-// NOTE: go-libp2p will be moving to dependency injection soon. The variadic
-// `extra` interface{} parameter facilitates the future migration. Supported
-// elements are:
-//  - connmgr.ConnectionGater
-func NewSwarm(local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, extra ...interface{}) *Swarm {
+func NewSwarm(local peer.ID, peers peerstore.Peerstore, bwc metrics.Reporter, opts ...Option) *Swarm {
```
Stebalien, line 110: The bandwidth is also optional. We can also make the peerstore optional, but then we'd need to make sure to close it iff we create it.

Stebalien: I'd also consider allowing this to return an error.

testing/testing.go
```diff
@@ -113,7 +113,11 @@ func GenSwarm(t *testing.T, opts ...Option) *swarm.Swarm {
 	ps.AddPrivKey(p.ID, p.PrivKey)
 	t.Cleanup(func() { ps.Close() })
 
-	s := swarm.NewSwarm(p.ID, ps, metrics.NewBandwidthCounter(), cfg.connectionGater)
+	var swarmOpts []swarm.Option
+	if cfg.connectionGater != nil {
+		swarmOpts = append(swarmOpts, swarm.WithConnectionGater(cfg.connectionGater))
```
Stebalien, line 118: I'd change the test swarm config to store a set of swarm "options" instead of a `connectionGater`. We'd then have a `WithSwarmOption` (or something like that) option to pass swarm options directly.

Stebalien (approved):
Couple of comments, but none are blocking.

testing/testing.go
marten-seemann: I considered that, but unfortunately the upgrader also requires us to set the connection gater.

swarm.go
marten-seemann: I converted the `metrics.Reporter` into an option. It doesn't look like the peerstore is optional at the moment, so why don't we just keep it as a parameter on the constructor?

Stebalien: > so why don't we just keep it as a parameter on the constructor?

No real reason. Making it optional and constructing it internally may make things a bit easier for the user, but it's likely too annoying to be worth it.

testing/testing.go
Stebalien: Ah...

Stebalien (approved):

PR #292, opened 2021-11-09
release version 0.8.0

guseggert:


marten-seemann (approved):
It's a bit unfortunate that this is another (minor) version bump, as we've just released v0.7.0, but the semver rules are clear and this PR is correct.

Maybe we can hold off on merging this, just in case we need to get something else in (not sure about our release plans for go-ipfs)?

guseggert:
> Maybe we can hold off on merging this, just in case we need to get something else in (not sure about our release plans for go-ipfs)?

Unfortunately getting this new version released is blocking downstream plumbing since this is a widely-consumed module, and we need to finish the plumbing as quickly as possible to minimize disruption.

PR #300, opened 2021-12-14
remove dependency on go-addr-util

marten-seemann:
Depends on https://github.com/multiformats/go-multiaddr/pull/168 and https://github.com/multiformats/go-multiaddr/pull/170.
This will allow us to deprecate the go-addr-util repo.

### DIFF
--- a/swarm_dial.go
+++ b/swarm_dial.go
@@ -11,7 +11,6 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/transport"
 
-	addrutil "github.com/libp2p/go-addr-util"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
 )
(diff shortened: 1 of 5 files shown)


swarm_dial.go
```diff
@@ -649,11 +648,18 @@ func (s *Swarm) filterKnownUndialables(p peer.ID, addrs []ma.Multiaddr) []ma.Mul
 		}
 	}
 
-	return addrutil.FilterAddrs(addrs,
-		addrutil.SubtractFilter(ourAddrs...),
+	return ma.FilterAddrs(addrs,
+		func(addr ma.Multiaddr) bool {
```
vyzo, line 652: what happened to `SubtractFilter`? It would be nice to have it and use it here instead of duplicating its code.

marten-seemann: This was the only occurrence of `SubtractFilter`, and the plan was to remove that function. It allocates a `map`, so it will probably be advantageous to replace it by something that doesn't allocate: https://github.com/libp2p/go-addr-util/blob/f3e9655430934836b83ea4bdf38c8599e0b17128/filter.go#L7-L17.

vyzo: ok, fair enough.

vyzo (approved):

PR #301, opened 2021-12-19
speed up the dial tests

marten-seemann:
This brings down the execution time of our test suite from from 19s to 9s.

### DIFF
--- a/dial_test.go
+++ b/dial_test.go
@@ -82,11 +82,11 @@ func acceptAndHang(l net.Listener) {
 }
 
 func TestSimultDials(t *testing.T) {
-	// t.Skip("skipping for another test")
 	t.Parallel()
 
 	ctx := context.Background()
 	swarms := makeSwarms(t, 2, swarmt.OptDisableReuseport)
+	defer closeSwarms(swarms)
 
 	// connect everyone
 	{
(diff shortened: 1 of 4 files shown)


dial_test.go
```diff
@@ -439,39 +431,18 @@ func TestDialBackoffClears(t *testing.T) {
...
+	ifaceAddrs1, err := s2.InterfaceListenAddresses()
+	require.NoError(t, err)
 	s1.Peerstore().AddAddrs(s2.LocalPeer(), ifaceAddrs1, peerstore.PermanentAddrTTL)
 
-	if c, err := s1.DialPeer(ctx, s2.LocalPeer()); err == nil {
-		c.Close()
-		t.Log("backoffs are per address, not peer")
-	}
-
-	time.Sleep(BackoffBase)
```
marten-seemann, line 461: I'm not sure why were sleeping here. Removing this doesn't change anything about this test, so maybe the test case was broken in the first place?

vyzo: to clear thebackofd internally i believe

vyzo: but indeed doesnt seem to matter.

dial_test.go
```diff
@@ -558,14 +519,13 @@ func newSilentListener(t *testing.T) ([]ma.Multiaddr, net.Listener) {
 }
 
 func TestDialSimultaneousJoin(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.Background())
```
vyzo, line 561: why change this to background? A fault could leave dangling dials.
But then again it maybe doesnt matter.

marten-seemann: It can't. This test blocks until all 3 go routine have returned:
```go
// raise any errors from the previous goroutines
for i := 0; i < 3; i++ {
	require.NoError(t, <-errs)
}
```

vyzo (approved):

PR #302, opened 2021-12-19
add constructor options for timeout, stop using transport.DialTimeout

marten-seemann:


dial_test.go
```diff
@@ -161,8 +156,9 @@ func newSilentPeer(t *testing.T) (peer.ID, ma.Multiaddr, net.Listener) {
 func TestDialWait(t *testing.T) {
 	t.Parallel()
 
-	ctx := context.Background()
```
vyzo, line 164: again here, why are we deopping this context?

marten-seemann: Because it didn't do anything, it's a `context.Background()`.

vyzo: ok.

vyzo (commented):
how will this option get propagated?

marten-seemann:
Like this: https://github.com/libp2p/go-libp2p/pull/1271

vyzo (approved):
ok, looks good.

PR #305, opened 2021-12-21
increment active dial counter in dial worker loop

### LINKED ISSUE #291: Increment 'active' counter in swarm dial worker loop
Not sure where it should be incremented but likely around this case:

https://github.com/libp2p/go-libp2p-swarm/blob/87ff7aca8ae058f7fd7bd9ada6bb255fe1b0b3a8/swarm_dial.go#L476-L490

cc @vyzo

vyzo:
Fixes #291


vyzo:
follow up for tests in #306

marten-seemann (approved):

PR #308, opened 2021-12-30
add support for the resource manager

marten-seemann:


swarm.go
```diff
@@ -76,6 +76,13 @@ func WithDialTimeoutLocal(t time.Duration) Option {
 	}
 }
 
+func WithResourceManager(m network.ResourceManager) Option {
```
vyzo, line 79: should this be required in the constructor instead?

vyzo (commented):
we need to Done the scopes on closure.

swarm.go
```diff
@@ -156,6 +165,9 @@ func NewSwarm(local peer.ID, peers peerstore.Peerstore, opts ...Option) (*Swarm,
 			return nil, err
 		}
 	}
+	if s.rcmgr == nil {
+		s.rcmgr = network.NullResourceManager
+	}
```
vyzo, line 170: I am a little concerned about this, if we forget the option we'll get a neutered swarm that does no resource management.

marten-seemann: It's a valid thing to have a swarm without resource management.
The only reason the transports are using a function parameter instead of an option is that we want to run constructor magic on them.

vyzo: ok, fair enough.

vyzo (approved):

PR #315, opened 2022-02-21
Refactor dial worker loop into an object and fix bug

### LINKED ISSUE #314: growing goroutine
More and more goroutines, more and more memory usage。

vyzo:
The dial worker loop has been refactored into an object to aid testability.
And more importantly, a goroutine leak is fixed by dropping the unnecessary active dial tracking logic.

Closes #314


dial_worker.go
```diff
@@ -78,8 +79,8 @@ func (w *dialWorker) loop() {
 	defer w.wg.Done()
 	defer w.s.limiter.clearAllPeerDials(w.peer)
 
-	triggerNow := make(chan struct{})
-	close(triggerNow)
+	ready := make(chan struct{})
```
marten-seemann, line 82: What does `ready` do? Why is is closed immediately? Can we have a comment plz?

dial_worker.go
```diff
@@ -57,9 +57,10 @@ type dialWorker struct {
 	nextDial    []ma.Multiaddr
 	triggerDial <-chan struct{}
 
+	complete <-chan struct{}
```
marten-seemann, line 60: Can we have a comment what this chan does?

dial_worker.go
vyzo: it's a channel signalling immediate action in the loop; but yes, can haz comment.

dial_worker.go
vyzo: sure!

vyzo: done.

dial_worker.go
vyzo: added comment.

vyzo:
squashed some commits to clean up history.

marten-seemann (approved):

